PR #10, opened 2014-08-20
[JENKINS-24374] - Added code to skip maven dependencies of type pom that were being added to the Libraries

eeb:
While working on changes to the [m2release plugin](https://github.com/jenkinsci/m2release-plugin), I found that the .hli file being produced by the maven-hpi-plugin contained paths to .pom files in the Libraries section. This only became an issue when I moved the m2release plugin to the 1.532 version of Jenkins core as the updated ClassPluginStrategy class used the AntClassLoader and no longer the URLClassloader to load dependencies. Attempting to start Jenkins using mvn hpi:run resulted in a java.util.zip.ZipException. 

Adding the following dependency from the m2release plugin to a pom reproduces the issue:

``` <dependency>
          <groupId>org.apache.maven.release</groupId>
          <artifactId>maven-release-manager</artifactId>
          <version>2.1</version>
          <exclusions>
              <exclusion>
                  <artifactId>maven-project</artifactId>
                  <groupId>org.apache.maven</groupId>
              </exclusion>
          </exclusions>
      </dependency>
```

To fix the issue I just skip dependencies of type pom.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
@@ -142,6 +142,10 @@ public class HplMojo extends AbstractHpiMojo {
                     continue OUTER;
             }
 
+            //Skip artifacts of type pom
+            if(artifact.getType().equals("pom"))
+                continue;
+
             ScopeArtifactFilter filter = new ScopeArtifactFilter(Artifact.SCOPE_RUNTIME);
             if (!artifact.isOptional() && filter.include(artifact.artifact)) {
                 paths.add(artifact.getFile().getPath());


oleg-nenashev:
I think a warning message is required

Prifiz:
Hello.
Was the described fix included to any of released plugin versions?
It becomes critical for our development process.

oleg-nenashev:
No, it has not been included :(
CC @jenkinsci/code-reviewers , we need feedback

👍 from me

Prifiz:
Is there any ability to accelerate this process?
We really need this fix.

oleg-nenashev:
Well, I'll ask people in IRC

daniel-beck:
Seems straightforward enough 👍 

A review by @stephenc can't hurt.

src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
```diff
@@ -142,6 +142,10 @@ private void buildLibraries(List<String> paths) throws IOException {
                     continue OUTER;
             }
 
+            //Skip artifacts of type pom
+            if(artifact.getType().equals("pom"))
```
batmat: In general IMO when comparing with a static operand it's always preferable to use the inverse form `"pom".equals(artifact.getType())` so that there's no chance of NPE (well, here if artifact is never null).

batmat:
:+1:

Prifiz:
I've already reproduced this bug by adding a couple of depends with type jar.
Maybe this case is also worth checking

stephenc:
Oh how I wish we fixed the plugin packing back in March 2011 when I originally suggested it... probably fine 👍 from the blind man on the galloping horse

jglick:
Sounds right on the face of it, but risk of regressions here is high; is this not something that could be covered by an integration test?

jtnord:
Obsoleted by #56 which adds the requested IT

PR #12, opened 2015-05-11
JENKINS-24827 Port hpi:run to jetty 8.x

Vlatombe:
This pull request replaces https://github.com/jenkinsci/maven-hpi-plugin/pull/11

The notation
````
<configuration>
  <contextPath></contextPath>
</configuration>
````
has been deprecated by the jetty-maven-plugin and is now exposed as
````
<configuration>
  <webApp>
    <contextPath></contextPath>
  </webApp>
</configuration>
````
So the plugins parent should be updated to use this new structure.

### CODE REVIEW

**Comment by Vlatombe** [comment]
The notation
````
<configuration>
  <contextPath></contextPath>
</configuration>
````
has been deprecated by the jetty-maven-plugin and is now exposed as
<configuration>
  <webApp>
    <contextPath></contextPath>
  </webApp>
</configuration>

So the plugins parent should be updated to use this new structure.

**Comment by Vlatombe** [comment]
Maybe you could look up if the pull request creator is part of the Cloudbees GH organization?

**Comment by stephenc** [comment]
@Vlatombe ha! that is what we were doing, but that was giving us a lot of false negatives. I suspect I am going to have to resort to a hard-coded list of bees :-(

**Comment by jtnord** [comment]
Given we are just soon to require servlet 3.x I wonder if this should go futher than jetty 8 which is only 3.0?

**Comment by Vlatombe** [comment]
I could give a shot to Jetty 9.x if this gets some traction.

File: src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java [hunk 1]
```diff
@@ -263,27 +261,18 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
 
         // set HUDSON_HOME
-        SystemProperty sp = new SystemProperty();
-        sp.setName("HUDSON_HOME");
-        sp.setValue(jenkinsHome.getAbsolutePath());
-        sp.setIfNotSetAlready();
+        setSystemPropertyIfEmpty("HUDSON_HOME",jenkinsHome.getAbsolutePath());
```
jtnord: should this also set JENKINS_HOME?

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -419,12 +408,16 @@ private void generateHpl() throws MojoExecutionException, MojoFailureException {
...
+        if (isExtractedWebAppDirStale(extractedWebAppDir, webAppFile)) {
             FileUtils.deleteDirectory(extractedWebAppDir);
         }
         
         super.configureWebApplication();
-        getWebApplication().setWebAppSrcDir(webApp);
+        getWebAppConfig().setWar(webAppFile.getCanonicalPath());
+        // cf. https://wiki.jenkins-ci.org/display/JENKINS/Jetty
+        HashLoginService hashLoginService = (new HashLoginService("Jenkins Realm"));
+        hashLoginService.setConfig(System.getProperty("jetty.home", ".") + "/etc/realm.properties");
```
jtnord: would this be better defaulted to `work/` or is `.`  that already our CWD?

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJettyMojo.java
```diff
@@ -1,5 +1,5 @@
 //========================================================================
-//$Id: AbstractJettyMojo.java 36040 2010-10-18 11:24:56Z kohsuke $
+//$Id$
```
jtnord: is this a complete copy of the upstream source or are there changes?

Vlatombe: The only change from upstream is the mojo annotations (javadoc -> java annotations)

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
Vlatombe: I doubt this would be used in practice anyway, but indeed this would be better set to `work`.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -262,28 +260,20 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
         }
 
-        // set HUDSON_HOME
-        SystemProperty sp = new SystemProperty();
-        sp.setName("HUDSON_HOME");
-        sp.setValue(jenkinsHome.getAbsolutePath());
-        sp.setIfNotSetAlready();
+        // set JENKINS_HOME
+        setSystemPropertyIfEmpty("HUDSON_HOME",jenkinsHome.getAbsolutePath());
+        setSystemPropertyIfEmpty("JENKINS_HOME",jenkinsHome.getAbsolutePath());
```
jtnord: this does break backwards compatibility with anyone who sets HUDSON_HOME and uses hpi:run - as JENKINS_HOME will take precedence over HUDSON_HOME (IIRC)
I'm inclined to say if you set one you set both - or only set JENKINS_HOME.

But we should be encouraging devs to use JENKINS_HOME as HUDSON is long dead :-)

jtnord:
:bee:

jglick:
:bee: without understanding most of the details. I have to assume you have tested this interactively with a variety of plugins, as there are a lot of special cases buried in this mojo that can regress easily, and there is zero test coverage.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
Vlatombe: I'll just remove HUDSON_HOME then.

PR #14, opened 2015-05-15
Make hpi:bundle-plugins fail if you forgot the <type> tag

stephenc:
@reviewbybees

### DIFF
--- a/src/it/bad-bundle/invoker.properties
+++ b/src/it/bad-bundle/invoker.properties
@@ -1,0 +1,21 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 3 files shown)


tfennelly:
:+1: 

Assuming it works ... don't see a test.

src/it/bad-bundle/invoker.properties
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+invoker.goals=clean generate-resources
+invoker.buildResult = failure
```
stephenc: @tfennelly here is the test

oleg-nenashev:
:+1:

jglick:
Oh, thank you, this will save a lot of headaches.

PR #15, opened 2015-05-15
Add the ability to bundle optional plugins separately from regular

stephenc:
*NOTE:* currently Jenkins core does not understand optionally bundled plugins, so either a module or a plugin would need to extract them somehow

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/BundlePluginsMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/BundlePluginsMojo.java
@@ -160,7 +160,7 @@ public class BundlePluginsMojo extends AbstractJenkinsMojo {
             // so much simpler to force the user to explicitly resolve the conflict by adding an explicit
             // dependency
             
-            Set<String> optionalDepenencyIssue = new LinkedHashSet<String>();
+            Set<String> optionalDependencyIssue = new LinkedHashSet<String>();
             for (MavenArtifact a: selected.values()) {
                 try {
                     final MavenProject pom = a.resolvePom();
@@ -174,7 +174,7 @@ public class BundlePluginsMojo extends AbstractJenkinsMojo {
                                     "%s: optional dependency of %s version %s conflicts with the bundled version %s",
                                     a.getArtifactId(), d.getArtifactId(), d.getVersion(), matching.getVersion());
                             getLog().error(message);
-                            optionalDepenencyIssue.add(message);
+                            optionalDependencyIssue.add(message);
                         }
                     }
                 } catch (ProjectBuildingException e) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/BundlePluginsMojo.java
```diff
@@ -115,17 +133,85 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                }
+            }
+
+            // now we just need to check if the optional dependencies have issues
+            // in theory we do not need to fail the build if a transitive dependency has issues
+            // we could just up-version it, but that would put us back to the start in terms of dependency resolution
+            // so much simpler to force the user to explicitly resolve the conflict by adding an explicit
+            // dependency
+            
+            Set<String> optionalDepenencyIssue = new LinkedHashSet<String>();
```
jglick: typo

jglick:
This feels like putting the cart before the horse. I see no explanation here of what the purpose of this feature is, and there is no PR offering support for it in core. Until the background is prepared, if you have some use case for this, it would be better to just run a forked version of the HPI plugin, or use some other postprocessing step.

src/main/java/org/jenkinsci/maven/plugins/hpi/BundlePluginsMojo.java
```diff
@@ -115,17 +133,85 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                }
+            }
 
             outputDirectory.mkdirs();
+            if (!nonOptional.containsAll(selected.keySet())) {
+                optionalOutputDirectory.mkdirs();
+            }
 
+            int artifactIdLength = "Artifact ID".length(); // how many chars does it take to print artifactId?
+            int groupIdLength = "Group ID".length(); // how many chars does it take to print artifactId?
```
tfennelly: Typo (at the end)

tfennelly:
LGTM insofar as I get what's happening. You and Jesse can decide on the justification.

stephenc:
@jglick 

Purpose: If you are building a custom distribution of Jenkins and bundling plugins, you may want to bundle additional plugins but not force them to be installed.

This change allows people building custom distributions to split out some plugins so that they are within the WAR file. Thus people running their custom Jenkins distribution on a machine that has no network connectivity will have the "optional" plugins and all mandatory dependencies of those plugins available on disk.

The packager of such a distribution will need to provide a mechanism to extract these "optional" bundled plugins separate from the mechanism in core. Most likely this will be using something like:

    ServletContext context = Jenkins.getInstance().servletContext;
    for( String path : Util.fixNull((Set<String>)context.getResourcePaths("/WEB-INF/optional-plugins"))) {
        // probably extract and dynamically install
    }

I intend to provide a jenkins-module to assist people creating custom distributions, but the API for it may require some thinking and I'd rather get the supporting changes (i.e. this) merged upstream beforehand.

jglick:
Well this PR seems useless until that `jenkins-module` has been created. And what if during its implementation you realize that the layout you chose here is not quite right? Seems more sensible to implement them together and propose them as a complete package (paired PRs). If for some reason you need the Maven plugin to be run in advance of the code which uses it, you can always cut an unofficial release from this branch as an interim step.

stephenc:
The only think I can envision changing is the default directory for optional plugin output. If I need to do anything else it would still be backwards compatible with this change

jglick:
:+1: on the change itself though I would still advise not merging it until there is a matching runtime module.

stephenc:
https://github.com/jenkinsci/optional-plugin-helper-module provides the runtime module

jglick:
Recommend you mention that module in the Javadoc for the new options.

PR #16, opened 2015-06-09
${basedir}/target/generated-test-sources/injected/

jglick:
For consistency with other Maven plugins. NetBeans will recognize it at this location, for example.

@reviewbybees


stephenc:
:+1:

tfennelly:
:+1:

PR #17, opened 2015-06-10
[JENKINS-28408] fixed support for jpi packaging type

daspilker:
I tried to add a test, but I run into problems and none of the existing tests run to up to the package phase, so I could not copy/extend a test.


oleg-nenashev:
LGTM

PR #18, opened 2015-06-17
[JENKINS-28932] Added warLibExcludes parameter

oleg-nenashev:
The comma-separated list of library files, which should not be copied to the target HPI file. Supported formats: jar, ejb, ejb-client, tld, par.

Purpose of the change: force the exclusion of JAR files, which are known to be completely shaded and relocated.

https://issues.jenkins-ci.org/browse/JENKINS-28933

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -68,6 +68,7 @@ import java.util.List;
 import java.util.Properties;
 import java.util.Set;
 import java.util.Collection;
+import java.util.TreeSet;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
@@ -184,6 +185,16 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
      */
     @Parameter
     private String dependentWarExcludes;
+    
+    /**
+     * The comma-separated list of library files, which should not be copied to
+     * the target HPI file.
+     * The following file types will be processed: jar, ejb, ejb-client, tld, par. 
+     * A file must be specified as a full name without a path (e.g. docker-java-1.3.0.jar)
+     * @since TODO
+     */
+    @Parameter
+    private String warLibExcludes;
 
     /**
      * [ws|tab|CR|LF]+ separated list of package prefixes that your plugin doesn't want to see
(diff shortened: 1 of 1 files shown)


KostyaSha:
Could you investigate how 'war' packaging works in this case?
@aheritier Isn't it continuation of story "unhardcode jar packaging and support shade"?

aheritier:
@KostyaSha Yes I think it is

oleg-nenashev:
I doubt we will be able to change the default behavior due to the risk of missing libs if "plugin B takes the dependency from another plugin declaring the unused dependency".

@KostyaSha @aheritier 
Is it possible to consider the manual contents management as a temporary solution?

recena:
:+1: But I would use `hpiArtifactExcludes` instead of `warLibExcludes`

oleg-nenashev:
> But I would use hpiArtifactExcludes instead of warLibExcludes

All other parameters have the same prefix, so I think it's OK

stephenc:
Favour a list of things to pass to artifact filters than a comma separated string... only use comma separated strings when dealing with command line properties. this is something we expect to live in the pom config

oleg-nenashev:
@KostyaSha 
Investigated maven-war-plugin. They use a similar approach, but a relative path is being used

```
<project>
  ...
  <build>
    <plugins>
      <plugin>
        <artifactId>maven-war-plugin</artifactId>
        <version>2.6</version>
        <configuration>
          <!--
            Exclude JCL and LOG4J since all logging should go through SLF4J.
            Note that we're excluding log4j-<version>.jar but keeping
            log4j-over-slf4j-<version>.jar
          -->
          <packagingExcludes>
            WEB-INF/lib/commons-logging-*.jar,
            %regex[WEB-INF/lib/log4j-(?!over-slf4j).*.jar]
          </packagingExcludes>
        </configuration>
      </plugin>
    </plugins>
  </build>
  ...
</project>
```

stephenc:
maven artifact filters has some helper methods so you could use a list of strings that get parsed by artifact filters

oleg-nenashev:
Seems a copy-pasting from maven-war-plugin is the best solution

recena:
@oleg-nenashev Then, the best option is `packagingExcludes`;)

KostyaSha:
One more thing that i found https://groups.google.com/forum/#!topic/jenkinsci-dev/8x2nrUwHg60

oleg-nenashev:
Does not make much sense anymore

PR #21, opened 2015-08-20
[JENKINS-30067] Only lookup annotations under classes/*

Vlatombe:
[JENKINS-30067](https://issues.jenkins-ci.org/browse/JENKINS-30067)

Not in any of the jars under WEB-INF/lib causing delays.

@reviewbybees


jglick:
A blind :bee:. Could you reproduce the problem, and does this seem to solve it?

Vlatombe:
Yes, I reproduced the problem on workflow-plugin before working on this.

PR #23, opened 2015-09-15
Maven 3 upgrade

jglick:
@stephenc do you know of any guide for replacing Maven 2.x API calls? I am finding it close to impenetrable.

Motivation: I found a plugin whose `dependency:tree` shows the expected version of a plugin included as a transitive dependency, yet `hpi:run` copies an older—and incompatible—version to `work/plugins/`. I suspect this is due to `RunMojo` calling 2.x APIs rather than Aether to find what to pass to `copyPlugin`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
     <openbracket>{</openbracket>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.level>6</java.level>
+    <maven.version>3.0.5</maven.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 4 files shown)


stephenc:
@rfscholte have you any advice... I seem to recall some adapter lib you wrote

rfscholte:
https://cwiki.apache.org/confluence/display/MAVEN/Plugin+migration+to+Maven3+dependencies describes the steps I've done so far.
Regarding maven-artifact-transfer, the API is not yet final,it might even get a new artifactId if we can think of name better covering its purpose.

KostyaSha:
Maybe rewrote this plugin from scratch?

stephenc:
Been threatening to do so since March 2nd 2011... just keep getting distracted

basil:
Closing in favor of #247

PR #24, opened 2015-10-23
[FIXED JENKINS-31133] Allow processing the plugin jar file before generating the hpi file

stephenc:
[JENKINS-31133](https://issues.jenkins-ci.org/browse/JENKINS-31133)

- The attached integration test verifies the jarsigner usecase (although manual verification was required as unzipping the .hpi to unzip the .jar to verify the signature present has been left as an excercise to the reader)
- The shade usecase reveals that the maven-shade-plugin does not currently support the designed use-case... off to maven land it would seem... (though the required hooks are present in this plugin)

@reviewbybees


jglick:
How does this relate to #18?

stephenc:
I think that #18 can be replaced by just marking those dependencies as `<optional>true</optional>` ;-) as that will exclude them from the `.hpi` file for non-`.hpi` dependencies... i.e. I don't think #18 is needed at all

src/it/process-jar/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+
+  <name>MyNewPlugin</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.6.3</version>
+      <!--TODO we mark it as optional to exclude from the WEB-INF/lib but when shade is fixed need to verify that it shades in these optional dependencies-->
+      <optional>true</optional>
```
stephenc: @jglick wrt #18 this marks the jackson dependencies as excluded from the .hpi file... in fact they will not be on the classpath at all... now there are a couple of issues with the current shade plugin:

* it cannot process a side artifact, it will only process the main artifact
* I am unsure if it processes optional artifacts (if it does then that should provide a workaround for @KostyaSha's issue in https://github.com/jenkinsci/maven-hpi-plugin/pull/18#issuecomment-126306038 as he can then just mark the dependency as `<optional>true</optional>` and stop trying to create a dependency reduced pom)

The two above can probably be fixed by me in the shade plugin...

For #18, I think that was assuming that you would use the maven-jar-plugin to generate the `jar` (which would then be the main artifact and then maven will blow up in your face when the hpi plugin tries to replace the main artifact with the hpi file... never mind that the .jar file resolved from the repository wouldn't be the shaded one at all)

So in short I think #18 was misguided and should probably be dropped. /cc @oleg-nenashev

src/it/process-jar/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+  <packaging>hpi</packaging>
+
+  <name>MyNewPlugin</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <version>2.6.3</version>
+      <!--TODO we mark it as optional to exclude from the WEB-INF/lib but when shade is fixed need to verify that it shades in these optional dependencies-->
```
jglick: Would you not use `<scope>provided</scope>` for that purpose?

src/it/process-jar/verify.groovy
```diff
@@ -0,0 +1,39 @@
...
+assert !new File(basedir, 'target/process-jar/WEB-INF/lib/jackson-databind-2.6.3').exists();
+assert new File(basedir, 'target/process-jar.hpi').exists();
+assert new File(basedir, 'target/process-jar.jar').exists();
+
+assert new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools/hpi/its/Messages.java').exists();
+
+content = new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools/hpi/its/Messages.java').text;
+assert content.contains(" holder.format(\"it.msg\");");
+
+// TODO add some test on hpi file content
```
jglick: So with a fix to the Shade plugin, the IT would be to verify that the Jackson lib is shaded and that the plugin classes are using it that way, right?

In the meantime, your IT seems to be running jarsigner, but :bug: it is not verifying that `target/process-jar.hpi!/WEB-INF/lib/process-jar.jar!/META-INF/KEY.SF` exists.

jglick: (unless I am missing something; much of this is beyond my ken)

stephenc: Yes I manually verified that but couldn't determine the correct groovy scripts to test the content... so `KEY.SF` is there in the `.jar` in the `.hpi`

stephenc: Yes with a fix to the Shade plugin we should be able to verify that the classes in `process-jar.hpi` have been shaded with the minimized Jar dependencies on jackson relocated into the shaded namespace and the whole lot signed

stephenc: Note that the TODO on the content tests was copied from the `verify` it that also doesn't test the content ;-)

src/it/process-jar/pom.xml
stephenc: So shade definitely doesn't process `<scope>provided</scope>` (by design from what I can tell)

src/it/process-jar/pom.xml
oleg-nenashev: #18 may be still useful, because it allows to exclude non-JAR stuff if there's such need. But I agree that shading use-cases could be solved in much more elegant way.

stephenc: The `<optional>true</optional>` solution applies to any dependency that is
not a `<type>hpi</type>` so I do not see any remaining use-case for #18​

src/it/process-jar/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <!--<configuration>-->
+              <!--<jarClassifier>unshaded</jarClassifier>-->
+            <!--</configuration>-->
+          </execution>
+        </executions>
+        <configuration>
+          <jarClassifier>shaded</jarClassifier>
+        </configuration>
+      </plugin>
+      <!-- TODO the shade plugin doesn't like shading the non-main attached artifacts yet -->
```
KostyaSha: AFAIR it wasn't executed at all, because hpi plugin has own cycle and jar archiver.

stephenc: Note I am discussing the way shade should be used not the crazy hacks around crazy hacks that people have been resorting to using.

With this change and a suitable release of the shade plugin then you could have the docker-plugin be a single module again without the need for docker-java-shaded at all

stephenc:
@jglick should be all happy now ;-)

src/it/process-jar/pom.xml
KostyaSha: @stephenc 1) i'm not maintaining docker-plugin anymore. 2) i resolved in upstream ability to produce already shaded artifact :) 3) when i debugged issue i found that hpi plugin directly creating jar archive renamed to hpi without calling real lifecycles (so shade plugin wasn't executed at all) plus maven-3.3 that picked wrong artifacts. 

I thought that `hpi` plugin should work like `war` without doing custom hardcoded magic.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -116,17 +117,44 @@ private void performPackaging()
...
+            jarFile = getOutputFile(".jar");
+            getLog().info("Generating jar " + jarFile.getAbsolutePath());
+            MavenArchiver archiver = new MavenArchiver();
+            archiver.setArchiver(jarArchiver);
+            archiver.setOutputFile(jarFile);
+            jarArchiver.addConfiguredManifest(manifest);
+            jarArchiver.addDirectory(getClassesDirectory());
+            archiver.createArchive(project, archive);
+        }
+        // HACK Alert... due to how this plugin hacks the maven dependency model (by using a dependency on the
```
KostyaSha: ooo yeah...

src/it/process-jar/pom.xml
stephenc: @KostyaSha w.r.t. 3

What is actually happening is that when you add `jar:jar` to the lifecycle (which you currently need in order to let `shade:shade` work) then the resulting `jar` file is attached as the main artifact... and then attempted to be replaced as the main artifact by `shade:shade`... then along comes `hpi:hpi` which produces its own `.jar` and tries to set the main artifact to the `.hpi` file...

Depending on who sets the main artifact first and which version of Maven you are running you can thus end up with any of three different artifacts as the main artifact, plus the `.jar` that `hpi:hpi` produced being added as a side attached artifact...

There is a secondary issue with producing the shaded jar file in the same reactor as consuming it when running Maven 3.3.x and where you create a dependency reduced pom... the work-around for that issue is to make the pom such that it is "effectively" dependency reduced already by exploiting `<optional>true</optional>` for all the dependencies that you want to remove

KostyaSha: omg... interesting what happens in gradle. Probably they have no such problem because they have no shaded support :laughing:

jglick:
:bee:

src/it/process-jar/pom.xml
KostyaSha: @stephenc what is the main artifact in `war` then? Why jenkins needs producing `jar` as main artifact and not `hpi`?

KostyaSha:
:+1: but i still don't understand why hpi need to be so crazy

src/it/process-jar/pom.xml
stephenc: https://github.com/jenkinsci/docker-plugin/pull/343 shows how to fix the current maven 3.3.x issues in docker-plugin BTW

stephenc: @KostyaSha now we get to the reason why I put in the hack-alert...

So how a hpi plugin packaging *should* work is that you declare the dependency as 

```
<dependency>
  <groupId>...</groupId>
  <artifactId>...</artifactId>
  <type>hpi</type>
  <version>...</version>
</dependency>
```

The key part here is the `<type>hpi</type>`

What instead the hpi plugin does is a super painful hack... it deploys a side artifact that is of type `jar` so that when you do 

```
<dependency>
  <groupId>...</groupId>
  <artifactId>...</artifactId>
  <version>...</version>
</dependency>
```

that (due to the default value of `type` in the maven schema) becomes

```
<dependency>
  <groupId>...</groupId>
  <artifactId>...</artifactId>
  <type>jar</type>
  <version>...</version>
</dependency>
```

Which means that the jar file is added to the classpath by surefire, etc... and all your tests work...

Now in Maven land, what we really want you to do with a `.war` file is not have any java code in there and use a multi-module build to produce the `.jar` file and then just consume that in the `.war` file... if we did that for Jenkins plugins then every Jenkins plugin would need to be three modules, one for the `.jar`, one for the `.hpi` and the last one as a parent (or we could go with my suggestion of a different plugin format that is actually classpath compatible and bundles the web stuff inside the `.jar` under e.g. `META-INF/web`)

So then to unwind the above hack, the hpi plugin parses the dependencies and adds manifest entries for any that have the corresponding pom file with `<packaging>hpi</packaging>` and leaves all the rest as being bundled in the `WEB-INF/lib` unless they are `<optional>true</optional>`

So yeah crazy layer of hack on hack on hack

stephenc:
@KostyaSha It has to be so crazy because the original author does not understand Maven

src/it/process-jar/pom.xml
KostyaSha: Can `hpi` somehow reorder `packaging` phase to be always latest?

KostyaSha:
@stephenc and how `maven-plugin` packaging works? Does jenkins really needs `hpi` file extension?

stephenc:
@reviewbybees done

stephenc:
This change has been sitting for 4 days... merge time!

jtnord:
REF maven shade issue [MSHADE-154](https://issues.apache.org/jira/browse/MSHADE-154)

PR #25, opened 2015-10-29
Coördinated run across repositories

jglick:
For those of us who routinely make coördinated changes across several source repositories, using snapshot dependencies while developing, it is very painful to have to rebuild artifacts every time we need to change something upstream.

With this patch, as soon as you declare a snapshot dependency on an upstream plugin or Jenkins core (https://github.com/jenkinsci/jenkins/issues/1894), you get the developer-friendly features of `hpi:run` that would otherwise be restricted to code in the plugin being run:

* Changes to `src/main/resources/` will be picked up immediately after saving the file and reloading the web page.
* Changes to `target/classes/` will be picked up after compiling, if the class has not yet been loaded. Typically, your IDE would implement “compile on save” and then JPDA hot reload can work.

`test`-scoped plugin dependencies also work for this purpose, if you only want to check interactions with another plugin, but not actually depend on it.

The trick is to keep an index of upstream Maven projects and the snapshot version that was last built from them. @stephenc yes I tried to make this work using the Maven reactor, but it just will not fly. Running

    mvn -f reactor/main-plugin hpi:run

does not allow you to resolve the source locations of sibling modules from the reactor; `Artifact.file` will be from `~/.m2/repository/`. Running

    mvn -f reactor -pl main-plugin hpi:run

sounds like it would fix that, but it does not. Running

    mvn -f reactor -pl api-plugin,main-plugin hpi:run

(or I guess `-f reactor -pl main-plugin -am` or just `-f reactor`) _does_ give you `reactor/api-plugin/target/classes/`, but then you are running the `run` goal on everything, which of course does not work.

[MNG-5059](https://issues.apache.org/jira/browse/MNG-5059) would seem to be the only way out of this, but it seems to have no traction. Anyway the current solution is simple enough, and relieves the developer of the need to create an artificial reactor POM just to try some matching changes.

@reviewbybees


stephenc:
I was saying to run something like

```mvn hpi:run -Dhpi.project=(groupId:)artifactId```

Where it would only start jetty in the module with the matching artifactId (with optional groupId in case of conflicts) and then you get the full reactor able to build.

**Comment by jglick** [comment]
Well, can try something like that if you think it would be the right way. (No need to pass `groupId`—we only care about plugins, and they are uniquely identified by `artifactId`.) Might need to specify `requiresDependencyCollection` or `requiresDependencyResolution` (not clear on the difference).

**Comment by abayer** [comment]
:bee: - this will be handy.

**Comment by jglick** [comment]
When I get a chance I will try @stephenc’s preferred approach.

**Comment by amuniz** [comment]
:bee: Nice feature

**Comment by amuniz** [comment]
I think @stephenc's proposal is going to work only for multi-module maven projects, but not with *any* snapshot dependency as @jglick described initially.

This can be merged as it is IMO and we all will be really happy (with the risk of having @stephenc being a :sad_panda: :smile_cat:

**Comment by jglick** [comment]
As I am working on sets of plugins with constantly shifting snapshot dependencies, being forced to create a new reactor project every time I want to try some changes interactively seems onerous. Much less overhead to `install` the dependencies I need (if I have not done so already) and `hpi:run` the most downstream. So I would like the original approach to be reconsidered unless there is some reason this would break things or lead to troubling side effects.

**Comment by andresrc** [comment]
:bee:

**Comment by michaelneale** [comment]
Yes pls.

File: src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java [hunk 1]
```diff
@@ -429,6 +442,16 @@ public void configureWebApplication() throws Exception {
         
         super.configureWebApplication();
         getWebAppConfig().setWar(webAppFile.getCanonicalPath());
+        for (Artifact a : (Set<Artifact>) project.getArtifacts()) {
+            if (a.getGroupId().equals("org.jenkins-ci.main") && a.getArtifactId().equals("jenkins-core")) {
+                File coreBasedir = TestHplMojo.readMap(a.getId());
+                if (coreBasedir != null) {
+                    List<File> additions = Arrays.asList(new File(coreBasedir, "src/main/resources"), new File(coreBasedir, "target/classes"));
+                    getLog().info("Will load directly from " + additions);
+                    getWebAppConfig().setWebInfLib(additions);
```
jglick: Does not compile after #28 incompatibilities.

kohsuke:
:bee: for this much needed change!

I added one more change to turn this map into a Plexus component. IMHO that's a better structure. See https://github.com/jenkinsci/maven-hpi-plugin/tree/coordinated-run in particular https://github.com/jenkinsci/maven-hpi-plugin/commit/0775329e8850e3ca2518627e5403b29b2ee6d63f

This should be also documented at https://wiki.jenkins-ci.org/display/JENKINS/Extend+Jenkins or a similar place.

jglick:
OK, will take me some time to retest after pulling that in.

oleg-nenashev:
@jglick any short-term plans regarding this PR

jglick:
Well I was ready to merge until @kohsuke changed the implementation, and I have not had time to retest. Are you waiting for it?

oleg-nenashev:
nope. Just wonder

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMapImpl.java
```diff
@@ -0,0 +1,70 @@
...
+    public PluginWorkspaceMapImpl(File mapFile) {
+        this.mapFile = mapFile;
+    }
+
+    public PluginWorkspaceMapImpl() {
+        this.mapFile = new File(System.getProperty("user.home"), ".jenkins-hpl-map");
+    }
+
+    private Properties loadMap() throws IOException {
+        Properties p = new Properties();
```
jglick: I know, it shows my age that I am not using JSON…

pom.xml
```diff
@@ -320,6 +326,18 @@
           <target>1.${java.level}</target>
         </configuration>
       </plugin>
```
oleg-nenashev: maven-compiler-plugin can be removed. It's added to 1.37

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMap.java
```diff
@@ -0,0 +1,19 @@
...
+ * the local workspaces of plugins under development.
+ *
+ * <p>
+ * The contract offers read/write access to a virtual {@code String -> File} map.
+ *
+ * @author Jesse Glick
+ * @author Kohsuke Kawaguchi
+ */
+public interface PluginWorkspaceMap {
+    File read(String id) throws IOException;
```
oleg-nenashev: 🐜 Add Javadoc and annotations ?

oleg-nenashev: And maybe since to this method and to others

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMapImpl.java
```diff
@@ -0,0 +1,70 @@
...
+        if (mapFile.isFile()) {
+            InputStream is = new FileInputStream(mapFile);
```
oleg-nenashev: Use Java 7 try-with-resource ?

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMapImpl.java
svanoort: Hah, but I'm with you there -- JSON is not a magic bullet for All The Things

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMap.java
oleg-nenashev: Especially since it's CheckForNull in the implementation...

abayer:
I refuse to approve any PR with a diaeresis in the title. =)

(kidding!)

Would this provide any benefit in PR CI unit test builds? I assume not, but that's something I'd really love to have.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -443,6 +458,16 @@ public void configureWebApplication() throws Exception {
         
         super.configureWebApplication();
         getWebAppConfig().setWar(webAppFile.getCanonicalPath());
+        for (Artifact a : (Set<Artifact>) project.getArtifacts()) {
+            if (a.getGroupId().equals("org.jenkins-ci.main") && a.getArtifactId().equals("jenkins-core")) {
```
oleg-nenashev: IIRC not all modules are in these packages

svanoort:
I think a 🐝  but I don't entirely understand the implication of this (i.e. there may be second-order impacts)

oleg-nenashev:
:bee:, but not a :happypanda:

jglick:
> Would this provide any benefit in PR CI unit test builds?

Sorry, this has no effect on tests, only `hpi:run`.

Someone, perhaps @stephenc, did propose (somewhere?) that if you file PRs with matching branches across repositories, the upstream PRs would get automatically `deploy`ed and then some Nexus tricks would kick in which force downstream builds to pick up the “right” snapshot dependencies. Not sure how that would work, but anyway if it does get implemented, it would be straightforward enough (I think) to amend this to categorize snapshots by branch name.

abayer:
'k, fair enough. :bee: it is then.

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMap.java
jglick: It is used only internally, there is no public API here.

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMapImpl.java
jglick: Could, yes.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
jglick: This is only purporting to handle changes made in `jenkins-core`. Modules could perhaps be added later if there is demand.

jglick:
Wrote up a note [here](https://wiki.jenkins-ci.org/display/JENKINS/Plugin+tutorial#Plugintutorial-DebuggingaPlugin).

src/main/java/org/jenkinsci/maven/plugins/hpi/PluginWorkspaceMapImpl.java
michaelneale: @jglick don't worry, in 2017 properties will be the new hotness file format. Trust me.

PR #27, opened 2016-01-14
[JENKINS-32493] Adapt to new plugin parent POM

andresrc:
[JENKINS-32493](https://issues.jenkins-ci.org/browse/JENKINS-32493)

@reviewbybees

### DIFF
--- a/hpi-archetype/pom.xml
+++ b/hpi-archetype/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.2</version>
+    <version>2.3</version>
     <relativePath/>
   </parent>
   <!-- $ --><groupId>org.jenkins-ci.tools</groupId><!-- /$ -->
@@ -15,12 +15,13 @@
   <packaging>hpi</packaging>
 
   <properties>
-    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <!-- Baseline Jenkins version you use to build the plugin. Users must have this version or newer to run. -->
     <jenkins.version>1.625.3</jenkins.version>
     <!-- Java Level to use. Java 7 required when using core >= 1.612 -->
     <java.level>7</java.level>
-    <!-- Baseline Jenkins version you use to build and test the plugin. -->
-    <jenkins-test-harness.version>2.0</jenkins-test-harness.version>
+    <!-- Jenkins Test Harness version you use to test the plugin. -->
+    <!-- For Jenkins version >= 1.580.1 use JTH 2.x or higher. -->
+    <jenkins-test-harness.version>2.1</jenkins-test-harness.version>
     <!-- Other properties you may want to use:
          ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin..
          ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin.
--- a/hpi-archetype/pom.xml
+++ b/hpi-archetype/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.0</version>
+    <version>2.2</version>
     <relativePath/>
   </parent>
   <!-- $ --><groupId>org.jenkins-ci.tools</groupId><!-- /$ -->
--- a/hpi-archetype/pom.xml
+++ b/hpi-archetype/pom.xml
@@ -6,8 +6,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
-    <version>1.580.1</version>
+    <version>2.0</version>
     <relativePath/>
   </parent>
   <!-- $ --><groupId>org.jenkins-ci.tools</groupId><!-- /$ -->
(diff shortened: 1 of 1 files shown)


hpi-archetype/pom.xml
```diff
@@ -15,15 +14,34 @@
...
+  <properties>
+    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <jenkins.version>1.580.1</jenkins.version>
+    <!-- Java Level to use. Default is 7, but was only required by Jenkins >= 1.614. -->
+    <java.level>6</java.level>
+    <!-- Other properties you may want to use:
+         ~ jenkins-test-harness.version: The JTH version used to test plugin. The default is ${jenkins.version}.
+         ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin. Current default is 1.115.
+         ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin. Current default is 1.17
+         ~ java.level.test: The Java version to use to build the plugin tests. Current default is 7.
```
stephenc: Do not encourage different test JDK version from production code. Currently only NetBeans can handle and it inhibits contributors using either of the other two IDEs

hpi-archetype/pom.xml
```diff
@@ -40,18 +58,6 @@
     <url>http://github.com/jenkinsci/${project.artifactId}-plugin</url>
   </scm>
    -->
-  <repositories>
```
stephenc: Unless we are going to start publishing the new parent to central you will need to retain this to let people build a clean checkout without adding the repos to their settings.xml (never mind other side issues)

hpi-archetype/pom.xml
andresrc: By "not encouraging" you mean not "advertising" it in the comments, don't you? The switch could be kept as it seems to be useful for some plugins.

hpi-archetype/pom.xml
andresrc: An if deployed to central, the parent pom should not reference other repos, ergo, let's move this back to the archetype and to the tested plugins.

:+1:

andresrc: Well, better copy than move, the parent pom may need it itself.

hpi-archetype/pom.xml
stephenc: no I mean that it's default value should be `${java.level}` not 7.

What you have described in the comments is that the main code is Java 6 and the tests have Java 7... this is of course on the one hand handy as the tests can then use try with resources and multi-catch... but now you cannot run the tests using Java 6 so your Java 6 code coverage is 0% and you have to hope that there is no subtle JVM version bug introduced.

Internally - at cloudbees - a person whose surname rhymes with stick changed one of our plugins to have the tests compiled with Java 8 because the IDE they used to be a developer for happens to support different Java levels for tests from production code... making it very hard for other people to work on that plugin... and making it impossible to test that plugin with Java 7 (or even Java 6 at the time the change was made)... The experience of working with that code base leads me to strongly recommend having the test level almost always the same as the main level

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,33 @@
   <!-- $ --><version>1.0-SNAPSHOT</version><!-- /$ -->
   <packaging>hpi</packaging>
 
+  <properties>
+    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <jenkins.version>1.580.1</jenkins.version>
```
jglick: Should change the default to the current LTS, 1.625.3 as of this writing.

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,33 @@
...
+    <!-- Java Level to use. Default is 7, but was only required by Jenkins >= 1.614. -->
+    <java.level>6</java.level>
```
jglick: Should switch to 7 if changing default baseline to 1.625.x or newer.

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,33 @@
...
+    <!-- Other properties you may want to use:
+         ~ jenkins-test-harness.version: The JTH version used to test plugin. The default is ${jenkins.version}.
```
jglick: No, this should now be specified explicitly, defaulting to 2.0.

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,33 @@
...
+         ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin. Current default is 1.115.
+         ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin. Current default is 1.17
```
jglick: Best not to document these “current defaults” in comments where they will soon be out of date.

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,33 @@
...
+    -->
+  </properties>
+
   <name>TODO Plugin</name>
   <description>TODO</description>
   <url>https://wiki.jenkins-ci.org/display/JENKINS/TODO+Plugin</url>
+
+  <!-- The default licence for Jenkins OSS Plugins is MIT. Uncomment or substitute for the applicable one.
```
jglick: Was intentional for this to be uncommented in the _archetype_. I do not think it should be silently _inherited_ from the parent.

hpi-archetype/pom.xml
```diff
@@ -40,18 +57,33 @@
...
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
       <url>http://repo.jenkins-ci.org/public/</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
```
jglick: BTW is this actually necessary?

amuniz: Actually, there are SNAPSHOTs in that repository, so I guess it should enabled for it.

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,32 @@
   <!-- $ --><version>1.0-SNAPSHOT</version><!-- /$ -->
   <packaging>hpi</packaging>
 
+  <properties>
+    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <jenkins.version>1.625.3</jenkins.version>
+    <!-- Java Level to use. -->
```
jglick: Would be polite to mention when core started requiring 7. (1.612)

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,32 @@
   <!-- $ --><version>1.0-SNAPSHOT</version><!-- /$ -->
   <packaging>hpi</packaging>
 
+  <properties>
+    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <jenkins.version>1.625.3</jenkins.version>
+    <!-- Java Level to use. Java 7 required when using core >= 1.612 -->
+    <java.level>7</java.level>
+    <!-- Baseline Jenkins version you use to build and test the plugin. -->
+    <jenkins-test-harness.version>2.0</jenkins-test-harness.version>
```
jglick: …though https://github.com/jenkinsci/jenkins-test-harness/pull/2 is pretty critical.

jglick:
:bee:

andresrc:
@reviewbybees again :)

hpi-archetype/pom.xml
```diff
@@ -15,15 +14,32 @@
   <!-- $ --><version>1.0-SNAPSHOT</version><!-- /$ -->
   <packaging>hpi</packaging>
 
+  <properties>
+    <!-- Baseline Jenkins version you use to build and test the plugin. Users must have this version or newer to run. -->
+    <jenkins.version>1.625.3</jenkins.version>
+    <!-- Java Level to use. Java 7 required when using core >= 1.612 -->
+    <java.level>7</java.level>
+    <!-- Baseline Jenkins version you use to build and test the plugin. -->
```
amuniz: Baseline Jenkins version? Isn't it JTH version?

jglick:
re:bee:

amuniz:
:bee:

andresrc:
@reviewbybees done

PR #28, opened 2016-02-24
[JENKINS-23378] update to Jetty9.2

kohsuke:
`hpi:run` should run Jetty 9.2 to support Servlet 3.1.

I've fixed how we track upstream files from jetty-maven-plugin so that future merges like this will be more manageable.

The diff view in this PR mixes upstream changes in Jetty and local changes, which is useless. If you want to really understand the diff, look at `git log -U jetty9.2 '^incoming' '^master'`

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -450,16 +450,6 @@ public class RunMojo extends AbstractJettyMojo {
         HashLoginService hashLoginService = (new HashLoginService("Jenkins Realm"));
         hashLoginService.setConfig(System.getProperty("jetty.home", "work") + "/etc/realm.properties");
         getWebAppConfig().getSecurityHandler().setLoginService(hashLoginService);
-
-        // use a bigger buffer as Stapler traces can get pretty large on deeply nested URL
-        for (Connector con : server.getConnectors()) {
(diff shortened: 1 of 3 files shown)


kohsuke:
This version runs git-plugin OK, so it's not dead on arrival.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -612,19 +651,13 @@ public void addJars(org.eclipse.jetty.util.resource.Resource lib) {
...
-
-                return con;
-            }
-        };
+    @Override
+    public void startJetty() throws MojoExecutionException {
+        if (httpConnector==null && defaultPort!=0) {
+            httpConnector = new MavenServerConnector();
+            httpConnector.setPort(defaultPort);
+        }
```
tfennelly: Maybe you meant this to be something like...

if (httpConnector==null) {
    httpConnector = new MavenServerConnector();
}
if (defaultPort!=0) {
    httpConnector.setPort(defaultPort);
}

tfennelly:
:bee: from what I can see.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
kohsuke: This is as intended. As the javadoc of `defaultPort` states, if a connector is configured that'll take precedence.

PR #29, opened 2016-03-16
We must use runtime scope for work/plugins/$p.hpl

jglick:
Without this fix,

    mvn -f maven-plugin hpi:run

fails with

```
… hudson.PluginManager$1$3$2$1 reactOnCycle
SEVERE: found cycle in plugin dependencies: (root=Plugin:maven-plugin, deactivating all involved) Plugin:maven-plugin -> Plugin:promoted-builds -> Plugin:maven-plugin
```

Same for the problem identified in https://github.com/jenkinsci/junit-plugin/pull/31.

@reviewbybees esp. @andresrc, @emanuelez


rsandell:
:bee:

andresrc:
belated :bee:

PR #32, opened 2016-04-28
[FIX JENKINS-34496] maven-hpi-plugin injecting test for non hpi packged projects

tfennelly:
See [JENKINS-34496](https://issues.jenkins-ci.org/browse/JENKINS-34496)

@jenkinsci/code-reviewers (esp @jglick)

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
@@ -47,6 +47,11 @@ public class TestInsertionMojo extends AbstractJenkinsMojo {
     }
 
     public void execute() throws MojoExecutionException, MojoFailureException {
+        if (!project.getPackaging().equals("hpi")) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -47,6 +47,11 @@ private static String quote(String s) {
     }
 
     public void execute() throws MojoExecutionException, MojoFailureException {
+        if (!project.getPackaging().equals("hpi")) {
```
batmat: NIT: I prefer always put the static string on the left side to make NPE impossible
`if ( !"hpi".equals(project.getPackaging()) ) {`

batmat:
:+1:

oleg-nenashev:
🐝

stephenc:
👎 Better would be to fix the actual tests so they don't fail the build as otherwise we can have jelly files in jenkins-modules with missing PIs

PR #20, opened 2015-08-11
Implement SimpleBuildStep on sample builder

jglick:
See [discussion](https://groups.google.com/d/msg/jenkinsci-dev/0dqXhbjRhYs/AAY6RUn6AgAJ).

With this change, if you create the sample project, `hpi:run` it, and install _Workflow: Aggregator_, _Snippet Generator_ suggests

```groovy
node {
  step([$class: 'HelloWorldBuilder', name: 'Bob'])
}
```

(well, without the required `node` context but that is another matter) and this works.

@reviewbybees

PR #33, opened 2016-06-09
[FIXED JENKINS-35506] Added HelloWorldStep to archetype.

abayer:
[JENKINS-35506](https://issues.jenkins-ci.org/browse/JENKINS-35506)

cc @reviewbybees esp @jglick and @svanoort


hpi-archetype/pom.xml
```diff
@@ -69,14 +69,22 @@
...
     </pluginRepository>
   </pluginRepositories>
-  <!-- If you want to depend on other plugins:
   <dependencies>
+    <!-- Remove this dependency if you don't need a Pipeline step -->
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-step-api</artifactId>
+      <version>2.0</version>
+      <optional>true</optional>
```
jglick: :-1: on optional dependencies. This one adds no GUI or large library dependencies—it is intended to be cheap.

hpi-archetype/src/main/resources/HelloWorldStep/global.jelly
```diff
@@ -0,0 +1,21 @@
...
+    To determine where this tag is defined, first check the namespace URI,
+    and then look under $JENKINS/views/. For example, <f:section> is defined
+    in $JENKINS/views/lib/form/section.jelly.
+
+    It's also often useful to just check other similar scripts to see what
+    tags they use. Views are always organized according to its owner class,
+    so it should be straightforward to find them.
+  -->
+  <f:section title="Hello World Builder">
+    <f:entry title="French" field="useFrench"
```
jglick: Delete this, it will not be used anyway—there is only one global configuration option, which is already associated with the `Builder`.

jglick:
I am not sure this is a good idea. First of all, we should not clutter the hello-world archetype with a usage of every extension point in Jenkins. There are plenty of worthy APIs. I would be less opposed if there were a separate archetype for an *empty* plugin (just `pom.xml` and `index.jelly`), in which case we could have a series of archetypes for plugins doing certain kinds of things.

Second, as of #20 the hello-world plugin is _already_ Pipeline compatible! Adding a `Step` just confuses matters by duplicating options, and as of JENKINS-29922 adds no value.

abayer:
I'd disagree as to the value - having an example of a `Step` has some value in and of itself, just as a quick reference, and subtly encouraging plugin devs to write actual `Step`s rather than just being compatible with `step([[$class: ...]])` is, I think, a very good thing.

jglick:
I do not want people writing `Step`s just for the sake of it. If `SimpleBuildStep` works, you should use it. Less code to maintain to get something that works the same in any project type. At any rate, having a `Step` example is fine, but not in the _same plugin_ as one which already has a `SimpleBuildStep`—you need to choose one or the other. So these would better be separate archetypes.

jglick:
:-1:, we should instead add a `@Symbol` to the existing step.

abayer:
Works for me. Closing.

PR #36, opened 2016-07-25
[JENKINS-36923] Allow declaring dependency on future release of core

stephenc:
- To enable the ability to unbundle bouncycastle we need to be able to target an unreleased Jenkins version as the baseline for a plugin

@jenkinsci/code-reviewers @reviewbybees 

See [JENKINS-36923](https://issues.jenkins-ci.org/browse/JENKINS-36923)

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.maven.plugins.hpi;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.factory.ArtifactFactory;
 import org.apache.maven.artifact.metadata.ArtifactMetadataSource;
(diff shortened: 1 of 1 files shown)


Vlatombe:
🐝

stephenc:
See also https://github.com/jenkinsci/bouncycastle-api-plugin/pull/14

jglick:
IIUC this would solve an aggravating issue in the current `split-plugins-from-core` process:

* extract the code to a new plugin and commit to the plugin and core
* update `DETACHED_PLUGINS` in core to mention it
* set the Jenkins baseline version in the plugin to the _last_ release (which still did include the code since split)
* release the plugin as `1.0-beta-1` so it does not appear on the UC for users of that weekly
* include `1.0-beta-1` in `detached-plugins` in the WAR
* wait for the next weekly core release
* update the baseline version to that release, then cut `1.0` of the plugin

With this change, you would be able to skip some steps; the plugin would go immediately to `1.0` in the UC but depending on a version of core which is expected several days later. Right?

CC @alvarolobato w.r.t. https://github.com/jenkinsci/jenkins/pull/2352.

Anyway, :bee:

stephenc:
@jglick correct

stephenc:
@reviewbybees done

stephenc:
I don't see any good reason to hold this change back any longer

daniel-beck:
Could this also be used to write extra fragile "Compatible with old releases but supporting new APIs if present" code?

stephenc:
@daniel-beck no because if the override is newer than the core dependency version we use the core dependency version in order to prevent stupid things from happening

PR #39, opened 2016-08-08
Sort artifacts so that Plugin-Dependencies appears in a consistent order from build to build

jglick:
While trying to track down what might have gotten broken in a plugin snapshot build, I noticed that the order of `Plugin-Dependencies` was random. Making it deterministic.

(Note that currently the primary sort key is effectively `groupId`, which means that from the perspective of plugin `shortName`s the list is not lexicographically sorted, though it _is_ deterministic.)

@reviewbybees


stephenc:
:bee: as the order is irrelevant to the uberclassloader's effectively random load order - in visibility terms - of a plugins dependencies

jglick:
> the order is irrelevant

So I would hope—but want to rule out a possible explanation for a problem I am seeing.

stephenc:
Well it's possible the order may have a seeding effect on the class loader graph, but initialisation timing could always revert that iiuc

PR #40, opened 2016-10-04
Reload scripts

KostyaSha:
https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/jenkins/model/Jenkins.java#L4832

Seems groovy reload doesn't work without it.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -298,6 +298,8 @@ public class RunMojo extends AbstractJettyMojo {
         // enable view auto refreshing via stapler
         setSystemPropertyIfEmpty("stapler.jelly.noCache","true");
 
+        setSystemPropertyIfEmpty("hudson.script.noCache","true");
+
         List<Resource> res = getProject().getBuild().getResources();
         if(!res.isEmpty()) {
             // pick up the first one and use it


KostyaSha:
cc ~~findbugsman~~ @oleg-nenashev

jglick:
You forgot to use a [permalink](https://github.com/jenkinsci/jenkins/blob/2f19ef0f8cd7aa96e6341120b93d5522f873eab6/core/src/main/java/jenkins/model/Jenkins.java#L4836) (`Y` button in GitHub).

jglick:
Sounds like it would solve the issue (whatever that is exactly). This will have effect on more than just that, though. For example, people trying to test the effect of using `resURL` will no longer be able to. I think it is not a big deal.

KostyaSha:
> Seems groovy reload doesn't work without it.

But i not sure, sometimes resources doesn't reload after changes at all...

jglick:
Rather than completely turning off caching, which makes interacting with Jenkins in `hpi:run` mode painfully slow, I would like to see Jenkins actually check file timestamps and invalidate its cache intelligently.

KostyaSha:
Even with this arg changed groovy scripts i.e. in config-detail.groovy in OptionalJobProperty doesn't reload. They reload while there is error and then somewhere cached. 

@jglick do you have any idea what else to check?

oleg-nenashev:
Just to respond to the ping in Skype from @KostyaSha. I ignore mentions which I find insulting.

KostyaSha:
So, 
https://github.com/stapler/stapler/blob/7065a603a6dde83f015bd8d600e1692673d37edb/groovy/src/main/java/org/kohsuke/stapler/jelly/groovy/GroovierJellyScript.java#L41-L50
Loads groovy file from `target/classes/$class/som.groovy` when it available. When i delete it, it picks groovy file from `src/main/resources` as expected.

KostyaSha:
Then https://github.com/stapler/stapler/blob/7065a603a6dde83f015bd8d600e1692673d37edb/core/src/main/java/org/kohsuke/stapler/lang/KlassNavigator.java#L154  returs groovy from target even if URLClassLoader contains `src/main/resources` first in list.

jglick:
> returns groovy from target even if URLClassLoader contains `src/main/resources` first in list

Sounds like it may be a bug in Stapler. Note that the order would not matter if your IDE automatically copies files from `src/main/resources/` to `target/classes/` after modification, as I think NetBeans at least does.

Best to check behavior first without an IDE (using `mvn compile` to test propagation of changes).

KostyaSha:
> if your IDE automatically copies files from src/main/resources/ to target/classes/ after modification

Is it default NetBeans behaviour or you jenkins-plugin?

jglick:
Default behavior.

PR #43, opened 2016-11-11
WIP Attempt to get optional plugins into hpi:run

KostyaSha:
Doesn't work, but any comments how to get optional deps for deps of hpi:run'ed plugin are welcome


KostyaSha:
@jglick @stephenc do you have any ideas?

stephenc:
Oh that's going to end up a hack fest very quickly... you probably have duplicate GAs with different Vs from walking the whole tree to recover the optionals...

You'all have to do a depth priority collection and pick the highest version for non-unique artifacts (remembering that when one gets evicted because its not the latest then you also need to evict all that the evicted dependency pulled in)

(Which explains some of the logic of PDTs in my Maven 5.0.0 proposal I am working on)

KostyaSha:
Collect all highest plugin numbers... may work...

stephenc:
Until it doesn't... and if dependency trees change from version to version you'll pull in incorrect deps

jglick:
I do not recall ever having had this problem. Would it not suffice to add `test`-scoped dependencies on transitive deps as a workaround?

PR #44, opened 2016-11-13
[FIX JENKINS-20679] Define manifest header in the jpi file to capture minimum Java requirement

batmat:
@jglick as you were commenting on that ticket. @daniel-beck @rtyler as we discussed that quite some times I think.

@stephenc for your Maven-eyes ;-)

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -957,6 +957,11 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
         mainSection.addAttributeAndCheck(new Attribute("Hudson-Version",jv));
         mainSection.addAttributeAndCheck(new Attribute("Jenkins-Version",jv));
 
+	    String javaLevel = project.getProperties().getProperty("java.level");
+	    if (StringUtils.isNotBlank(javaLevel)) {
+		    mainSection.addAttributeAndCheck(new Attribute("Java-Level", javaLevel));
+	    }
+
         if(maskClasses!=null)
             mainSection.addAttributeAndCheck(new Attribute("Mask-Classes",maskClasses));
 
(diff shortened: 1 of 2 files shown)


batmat:
BTW, it's not perfect at all. People could have a java.level set to 7 and still override the compiler settings.
But, granted, doing that manually when there's just one prop to set is quite [unlikely](https://github.com/jenkinsci/docker-plugin/blob/master/pom.xml#L70L78).

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -957,6 +957,11 @@ protected void setAttributes(Section mainSection) throws MojoExecutionException,
         mainSection.addAttributeAndCheck(new Attribute("Hudson-Version",jv));
         mainSection.addAttributeAndCheck(new Attribute("Jenkins-Version",jv));
 
+	    String javaLevel = project.getProperties().getProperty("java.level");
```
daniel-beck: Mixed spaces and tabs on the same line.

daniel-beck:
This looks a lot like it comes to late to prevent a necessary update in the backend-update-center2. Still, doesn't look like it can't hurt.

Can't you get the actual compiler version used?

batmat:
Well, yes, we have it as `Build-Jdk`, but it does not say which source/target was used...

stephenc:
It would seem easiest to me to just use an animal-sniffer lite approach and scan all dependencies for the max bytecode version and publish that in the header (you can ignore hpi files with the header, and allow manual specification for the rare cases where someone who knows what they are doing)

daniel-beck:
> manual specification for the rare cases where someone who knows what they are doing

WHAT COULD POSSIBLY GO WRONG?

stephenc:
> WHAT COULD POSSIBLY GO WRONG?

Well suppose I compile some classes with Java 8 but use code not to hit those classes when running with Java 8.

The more critical issue IMHO is dependencies

batmat:
superseded by https://github.com/jenkinsci/maven-hpi-plugin/pull/75

PR #46, opened 2016-12-07
Parse Short-Name & Plugin-Version out of the actual plugin manifest

jglick:
Ignoring the GAV used to express the dependency. Seems to fix a problem in the contents of `Plugin-Dependencies` in `target/test-classes/the.hpl` and `target/*.hpi` when using jitpack.io as in https://github.com/jenkinsci/plugin-pom/pull/37.

Possibly other uses of dependency GAVs are also wrong, say in the `run` goal (?), but this is the one I noticed for now.

@reviewbybees


stephenc:
:bee:

PR #47, opened 2016-12-16
New Jenkinsfile

jglick:
@reviewbybees esp. @rtyler who should please create a ci.jenkins.io job for this, to replace [`Libraries/maven-hpi-plugin`](https://ci.jenkins.io/job/Libraries/job/maven-hpi-plugin/) using the Evil Job Type.

Once there is a home, the *Build Status* link in `README.md` could be updated (or just deleted—we expect `master` to always be a blue ball, right?).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,9 @@
+node('docker') {
+  checkout scm
+  docker.image('maven:3.3.9-jdk-8').inside {
+    sh 'mvn -B -Prun-its clean install'
+  }
+  /* Useful for debugging failures, but currently too big (~25Mb):
+  archiveArtifacts 'target/its/*/build.log'
+  */
+}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,4 @@
+properties([buildDiscarder(logRotator(numToKeepStr: '20'))])
 node('docker') {
   checkout scm
   docker.image('maven:3.3.9-jdk-8').inside {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,6 @@ node('docker') {
   docker.image('maven:3.3.9-jdk-8').inside {
     sh 'mvn -B -Prun-its clean install'
   }
-  /* Useful for debugging failures, but currently too big (~25Mb):
-  archiveArtifacts 'target/its/*/build.log'
-  */
+  // Useful for debugging failures, but currently too big (~25Mb):
+  // archiveArtifacts 'target/its/*/build.log'
 }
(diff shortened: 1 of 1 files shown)


KostyaSha:
> using the Evil Job Type.

Workflow

KostyaSha:
Where is this information in new "the best job type" https://ci.jenkins.io/job/Libraries/job/maven-hpi-plugin/15/org.jenkins-ci.tools$maven-hpi-plugin/executedMojos ?

jglick:
The log file.

KostyaSha:
> The log file.

Where do you see mojo execution time in log?

Jenkinsfile
```diff
@@ -0,0 +1,8 @@
+node('docker') {
+  checkout scm
+  docker.image('maven:3.3.9-jdk-8').inside {
+    sh 'mvn -B -Prun-its clean install'
+  }
+  // Useful for debugging failures, but currently too big (~25Mb):
+  // archiveArtifacts 'target/its/*/build.log'
```
rtyler: Is there any reason not to have `junit` here for test archiving/reporting?

jglick:
> Where do you see mojo execution time in log?

Who cares? If you do, ask the Maven team to include that information. Not Jenkins’ responsibility.

> Is there any reason not to have junit here for test archiving/reporting?

There are no JUnit tests in this component at the moment, so no Surefire reports either. The tests are all using Invoker, which has its own format we cannot easily parse. If a test fails, the build will fail.

jglick:
@rtyler can we get `maven-hpi-plugin` added to the inclusion list on [`Plugins`](https://ci.jenkins.io/job/Plugins/)? Or would you prefer to manually create a multibranch project elsewhere, since this is not a plugin per se?

KostyaSha:
> Who cares? If you do, ask the Maven team to include that information. Not Jenkins’ responsibility.

Just please don't say that non-workflow jobs are evil. You are switching informative builds to not informative without reasons.

Jenkinsfile
KostyaSha: Check build status and attach, what's the problem?

Jenkinsfile
```diff
@@ -0,0 +1,12 @@
+node('docker') {
```
rtyler: Neglected to mention earlier, I think this should discard old builds automatically, e.g. 


```groovy
properties([[$class: 'jenkins.model.BuildDiscarderProperty', strategy: [$class: 'LogRotator', numToKeepStr: '32']]])
````

jglick:
Filed [INFRA-1006](https://issues.jenkins-ci.org/browse/INFRA-1006) to track job creation.

Jenkinsfile
jglick: (resolved)

jglick:
@abayer feel free to suggest the Declarative equivalent, though we have had some problems recently with incompatible syntax changes breaking ci.jenkins.io jobs.

PR #49, opened 2017-01-22
fix injected test syntax and doc

evernat:


src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -108,15 +108,18 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             f.mkdirs();
             File javaFile = new File(f, injectedTestName + ".java");
             PrintWriter w = new PrintWriter(new OutputStreamWriter(new FileOutputStream(javaFile)));
-            w.println("import java.util.*;");
             w.println("/**");
             w.println(" * Entry point to auto-generated tests (generated by maven-hpi-plugin).");
-            w.println(" * If this fails to compile, you are probably using Hudson &lt; 1.327. If so, disable");
-            w.println(" * this code generation by configuring maven-hpi-plugin to &lt;disabledTestInjection&gt;true&lt;/disabledTestInjection&gt;.");
+            w.println(" * It's possible to disable this code generation by configuring maven-hpi-plugin to &lt;disabledTestInjection&gt;true&lt;/disabledTestInjection&gt;.");
```
jglick: use `@code`

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -108,15 +108,18 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             w.println(" */");
             w.println("public class " + injectedTestName + " extends junit.framework.TestCase {");
+            w.println("  /**");
+            w.println("   * Build test.");
+            w.println("   * @return test");
+            w.println("   * @throws Exception e");
+            w.println("   */");
             w.println("  public static junit.framework.Test suite() throws Exception {");
-            w.println("    Map parameters = new HashMap();");
+            w.println("    java.util.Map<String, String> parameters = new java.util.HashMap<>();");
```
jglick: I think this will break plugins still on `java.level=6`.

evernat:
Concerning breaking java.level=6: I have doubts that plugins still targeting java.level=6 for old Jenkins versions will use the latest maven-hpi-plugin 2.1. And I think that those plugins will certainly continue to use an old version of the maven-hpi-plugin.
Anyway, I have now added a check in order to not break the plugin's build if that is the case.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -102,22 +105,29 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             w.println("public class " + injectedTestName + " extends junit.framework.TestCase {");
+            w.println("  /**");
+            w.println("   * Build test.");
+            w.println("   * @return test");
+            w.println("   * @throws Exception e");
+            w.println("   */");
             w.println("  public static junit.framework.Test suite() throws Exception {");
             w.println("    System.out.println(\"Running tests for \"+" + quote(project.getGroupId() + ":" + project.getArtifactId() + ":" + project.getVersion())+");");
-            w.println("    Map parameters = new HashMap();");
+            w.println("    java.util.Map<String, String> parameters = new java.util.HashMap<>();");
```
jglick: merge conflict

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -102,22 +105,29 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             getLog().info("Skipping auto-test generation because we are targeting Jenkins "+target+" (at least 1.327 is required).");
             return;
         }
+        if (javaLevel != null && javaLevel.compareTo("7") < 0) {
```
jglick: This will break on Java 10 BTW; use a numeric comparison.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -102,22 +105,29 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             w.println("public class " + injectedTestName + " extends junit.framework.TestCase {");
+            w.println("  /**");
+            w.println("   * Build test.");
+            w.println("   * @return test");
+            w.println("   * @throws Exception e");
+            w.println("   */");
             w.println("  public static junit.framework.Test suite() throws Exception {");
             w.println("    System.out.println(\"Running tests for \"+" + quote(project.getGroupId() + ":" + project.getArtifactId() + ":" + project.getVersion())+");");
-            w.println("    Map<String, Object> parameters = new HashMap<String, Object>();");
+            w.println("    java.util.Map<String, Object> parameters = new java.util.HashMap<>();");
```
jglick: Is this really worth disabling Java 6 support for??

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -102,22 +105,29 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
             File f = new File(project.getBasedir(), "target/generated-test-sources/injected");
             f.mkdirs();
             File javaFile = new File(f, injectedTestName + ".java");
             PrintWriter w = new PrintWriter(new OutputStreamWriter(new FileOutputStream(javaFile)));
-            w.println("import java.util.*;");
             w.println("/**");
             w.println(" * Entry point to auto-generated tests (generated by maven-hpi-plugin).");
-            w.println(" * If this fails to compile, you are probably using Hudson &lt; 1.327. If so, disable");
-            w.println(" * this code generation by configuring maven-hpi-plugin to &lt;disabledTestInjection&gt;true&lt;/disabledTestInjection&gt;.");
+            w.println(" * It's possible to disable this code generation by configuring maven-hpi-plugin to {@code <disabledTestInjection>true</disabledTestInjection>}.");
```
daniel-beck: Why are we even explaining this, given what we know about the insane side effects of this?

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
evernat: I think that the jenkins plugins upgrading to a recent version of plugin parent and of maven hpi plugin can upgrade their java version to 1.7 at least. I see no reason they can't upgrade their java version at that moment: what are they waiting otherwise?
So my answer is yes, it's worth it.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
evernat: @daniel-beck 
I don't know the answer, but it may be in another PR. (I am eventually trying to make this PR merged since January.)

src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
evernat: done

jglick:
Obsolete I think.

We _do_ need to figure out whether this system can be converted to JUnit 4.

PR #51, opened 2017-02-17
It's OK not to have classes dir.

kohsuke:
... though rare. Previously, this would have caused 'no such directory' error during execution. With this change, now it creates an empty jar file.


kohsuke:
@reviewbybees

aheritier:
🐝

stephenc:
:ant: if we don't have a classes directory can you auto-generate index.jelly from the Pom description?

jglick:
Ah, good point. We should at a minimum print a warning if `target/classes/index.jelly` is missing—every plugin should have it.

I presume this is somehow related to https://github.com/jenkinsci/jenkins/pull/2754 but @kohsuke is not giving any context.

kohsuke:
l tried to generate `index.jelly` but packaging is a wrong moment to do it. jar has already been created and tests have already run.

I'll just put `index.jelly` manually instead of trying to be smart.

PR #52, opened 2017-02-27
Allow setting of context path

i386:
You cant deprecate `contextPath` for `webApp.contextPath` because `webApp` object gets created during the 'execute' method and when maven tries to set it, its null.

This PR just restores the functionality lost. If someone wants it to work the way it was previously advertised they should but this would get Blue Ocean devs out of a productivity hole.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -213,10 +213,8 @@ public class RunMojo extends AbstractJettyMojo {
     /**
      * The context path for the webapp. Defaults to the
      * name of the webapp's artifact.
-     *
-     * @deprecated Use &lt;webApp&gt;&lt;contextPath&gt; instead.
      */
-    @Parameter(readonly = true, required = true, defaultValue = "/${project.artifactId}")
+    @Parameter(required = true, defaultValue = "/${project.artifactId}")
     protected String contextPath;
 
     @Component
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -227,7 +227,6 @@ public class RunMojo extends AbstractJettyMojo {
 
         if (webApp == null || webApp.getContextPath() == null) {
             if (contextPath != null) {
-                getLog().warn("Please use `webApp/contextPath` configuration parameter in place of the deprecated `contextPath` parameter");
                 if (webApp == null) {
                     try {
                         webApp = new JettyWebAppContext();


jglick:
Seems to be amending #12 & #28.

Unfortunately `RunMojo` has no IT yet which could prove that this is fixing a bug and make sure it does not regress. Not sure exactly how it should work. We would want to have it start Jenkins, inspect some stuff about the loaded plugin, then `${contextPath}/exit`, but the verification Groovy script runs only after the Maven build completes, and the build will not complete until Jenkins exits. Maybe the plugin can shut down Jenkins itself after performing some self-tests?

stephenc:
So webApp is likely `null` because you didn't specify the implementation class, i.e.

```
<configuration>
  <webapp  implementation="org.eclipse.jetty.maven.plugin.JettyWebAppContext">
    <contextPath>...</contextPath>
    ...
```

*should* work... now there is another question as to why you need to specify the type when e.g. Jetty 9's maven plugin doesn't... most likely that is the result of further bad decisions taken many many years ago when this plugin was first created.

jglick:
> get Blue Ocean devs out of a productivity hole

BTW what is the motivation for this—why are you trying to override the context path? We picked a nonempty default after years of painful experience.

i386:
@jglick We've got the reverse problem - running with /Jenkins but not from / 

@stephenc thats not going to work from the command line 

I am not going to write a integration test harness for this change. I quite simply don't have the time and frankly we just need to be able to test with and without a context path ASAP.

jglick:
> We've got the reverse problem - running with /Jenkins but not from /

Meaning that you are fine with the default being `/jenkins` but _also_ do interactive testing with an empty context path? Is there some reason to believe B.O. code which works correctly with an arbitrary context path would be liable to fail with an empty one?

(Certainly there is plenty of reason to think that plugin code which works with an empty context path will fail with a nonempty one—this happens all the time, when people use the wrong API calls.)

i386:
@jglick that's right. We've got problems calculating back links to classic right now that didn't take into account that Jenkins was being served from '/' that 5 mins of interactive testing would have picked up :(

We can't trust Jenkins server URL being set correctly all the time so we have to take it into account

jglick:
> We've got problems calculating back links to classic right now that didn't take into account that Jenkins was being served from '/'

Now you have piqued my curiosity—I have never heard of a bug affecting _only_ empty context paths.

> We can't trust Jenkins server URL being set correctly all the time

Sure you can. You have to.

i386:
Anyway, this isn't what we need help with right now. We'd like to test with different context paths.

michaelneale:
@jglick just bad assumptions being baked in - given much development happens via hpi:run.

jglick:
> you didn't specify the implementation class

Probably no one needs to configure a context path in their POM. I think the PR is addressing

    mvn hpi:run -DcontextPath=

which does not in fact work.

michaelneale:
@jglick correct

stephenc:
Try my config suggestion but use a property to set the context path, then you can override via the cli

jglick:
Might be a momentary workaround, but we definitely want to be able to override via CLI without any POM edits.

FWIW I do not feel comfortable reviewing the code change, which is why I assigned the people who seemed to have changed this code in the past. I am keen on getting this mojo covered by ITs, especially since we seem to be seeing an unnoticed regression from earlier changes.

i386:
using @stephenc's suggested workaround https://github.com/jenkinsci/blueocean-plugin/pull/870 but sad that this tooling won't see an update anyway.

stephenc:
@i386 I am not against changing the plugin, I am against deviation from the behaviour of the jetty plugin this is based on. (I'd really prefer not to even need this plugin, and let the jetty plugin do it all)

I suggested the workaround as a shortcut for BO, though we really need to know *why* you need it as having a contextPath has caught and killed more bugs than using / as the context path

i386:
@stephenc we have to do all sorts of calculations based of off the base URL, so being able to switch between `/` and `/Jenkins` locally would have uncovered some issues.

michaelneale:
@stephenc yeah default is still to have a context path, just nice to have a quick option to set it to something else when reviewing - which seems satisfied for now.

i386:
Is there any reason why we don't want to merge this?

PR #53, opened 2017-03-06
[JENKINS-37619] Add link to all the plugin goals

jtnord:
Adds a link to all the plugins goals.

@reviewbybees

### DIFF
--- a/src/site/apt/index.apt
+++ b/src/site/apt/index.apt
@@ -8,7 +8,7 @@ Maven Jenkins Plugin
   This maven2 plugin provides various goals for developing Jenkins plugins.
 
 
-* Goals Overview
+* Common Goals Overview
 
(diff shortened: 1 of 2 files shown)


src/site/apt/index.apt
```diff
@@ -8,7 +8,7 @@ Maven Jenkins Plugin
   This maven2 plugin provides various goals for developing Jenkins plugins.
 
 
-* Goals Overview
+* Common Goals Overview
```
oleg-nenashev: Does it change the anchor link BTW?

jtnord: Yes - but there is nothing on this page so the anchor is mostly pointless

ref: https://jenkinsci.github.io/maven-hpi-plugin/

jtnord:
@jglick added site to the run (but no archive so it is really only a sanity check)

PR #55, opened 2017-03-18
cover the case when plugin.version.description property is set to empty string in pom.xml

builderall:
Thanks much. 
This has to do with https://github.com/jenkinsci/docker/blob/master/install-plugins.sh. That is how I found it. This script takes space separated list of plugins to automatically install during docker image build. And the maven hpi package adds a space if it is a SNAPSHOT version in MANIFEST.MF "(Plugin-Version: 2.0.5-SNAPSHOT (githash-user/timestamp)", so the scripts thinks there are two plugins to install and fails.  If one explicitly puts quotes around the plugin name it still works. But you have to call it separately from other plugin install. Otherway might be to add another maven property to remove the space in before "(". But the cureent property can be effectively used the way developer wants it.

<properties>

<plugin.version.description></plugin.version.description>
</properties>

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -942,7 +942,7 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
                 dt = new SimpleDateFormat("MM/dd/yyyy HH:mm").format(new Date());
             pluginVersionDescription = "private-"+dt+"-"+System.getProperty("user.name");
         }
-        if (pluginVersionDescription!=null)
+        if (pluginVersionDescription!=null && !pluginVersionDescription.trim().isEmpty())
             v += " (" + pluginVersionDescription + ")";
 
         if (!project.getPackaging().equals("jenkins-module")) {


jglick:
Why would `pom.xml` set this property? It is set automatically when doing a snapshot build, otherwise not.

jglick:
And even if there were a bogus `" ()"` string in the version, it sounds like a bug in the Docker image that it is not properly quoting arguments.

daniel-beck:
Isn't the use case just a regression introduced by careless handling of whitespace in https://github.com/jenkinsci/docker/pull/365?

builderall:
Use case: Please take a look at Jenkins docker image https://github.com/jenkinsci/docker/install-plugins.sh , it doesn't work with spaces. Also we cannot force the versioning on everyone.It should be customizable, or at least should make sure world uses less space :)  Thanks

builderall:
The main contention is a space  after : , Plugin-Version: 2.0.5-SNAPSHOT . The plug in downloader/installer script treats it as another plugin and tries downloading it. So giving a way to customize the way would be helpful. Thanks again.

daniel-beck:
@builderall Yes, that's basically what I'm saying. To me it looks like a bug in `install-plugins.sh` which should be fixed, rather than something worked around here.

builderall:
@daniel-beck, here is what I remember now. The script works for all published plug-ins as they do not have space after version info or this suffix "  (githash-user/timestamp)". So any local compile of jenkins plugins that do not have binaries available get a free space in version name because of hpi plug-in's behavior, so it would be more user friendly to give that option or remove the whitespace completely. And the shell scripts get complicated easily, but something could be done there too. But it would still be better to give choices to the end users. Thanks.

daniel-beck:
As per @jglick, has this ever been tested? Seems to me that this won't actually change anything with the version numbers described. It would need to look like `1.2.3 ()` (empty pair of parens) to make a difference.

jglick:
Whatever this was solving, it is likely obsolete.

PR #56, opened 2017-03-22
[JENKINS-24374] Do not add non pom artifacts to the hpl classpath

jtnord:
Includes and obsoletes #10
Fixes [JENKINS-24374](https://issues.jenkins-ci.org/browse/JENKINS-24374)

Made worse as this affects not only the hpi:run but the standard Injected tests as this runs the plugin as part of the InjectedTests (to check the plugin starts - which it will not if it has a dependency of type `pom`)

> While working on changes to the [m2release plugin](https://github.com/jenkinsci/m2release-plugin), I found that the .hli file being produced by the maven-hpi-plugin contained paths to .pom files in the Libraries section. This only became an issue when I moved the m2release plugin to the 1.532 version of Jenkins core as the updated ClassPluginStrategy class used the AntClassLoader and no longer the URLClassloader to load dependencies. Attempting to start Jenkins using mvn hpi:run resulted in a java.util.zip.ZipException. 
> 
> Adding the following dependency from the m2release plugin to a pom reproduces the issue:
> 
> ``` <dependency>
>           <groupId>org.apache.maven.release</groupId>
>           <artifactId>maven-release-manager</artifactId>
>           <version>2.1</version>
>           <exclusions>
>               <exclusion>
>                   <artifactId>maven-project</artifactId>
>                   <groupId>org.apache.maven</groupId>
>               </exclusion>
>           </exclusions>
>       </dependency>
> ```
> 
> To fix the issue I just skip dependencies of type pom.

@reviewbybees /cc @eeb

### DIFF
--- a/src/it/pom-deps/pom.xml
+++ b/src/it/pom-deps/pom.xml
@@ -1,0 +1,37 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>plugin</artifactId>
+      <version>2.25</version>
+    </parent>
+
+    <groupId>org.jenkins-ci.tools.hpi.its</groupId>
+    <artifactId>pom-deps</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>hpi</packaging>
+    
+    <properties>
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+      <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+      <hpi-plugin.version>@project.version@</hpi-plugin.version>
+      <enforcer.skip>true</enforcer.skip>
+      <jenkins.version>2.7</jenkins.version>
+    </properties>
+
+
+    <name>PluginWithPomDependencues</name>
+
+    <dependencies>
+      <dependency>
+        <!-- https://issues.jenkins-ci.org/browse/JENKINS-24374 -->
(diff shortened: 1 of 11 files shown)


jglick:
BTW the [IT fails](https://ci.jenkins.io/job/Tools/job/maven-hpi-plugin/job/PR-56/3/artifact/target/its/pom-deps/build.log). Hooray for the `Jenkinsfile`!

```
Running post-build script: /home/jenkins/workspace/ools_maven-hpi-plugin_PR-56-AORYQTUGQ4LXFULWYYTKNYE3PDFKEQLVGLEFYDAZMSYHMVBNRHHQ/target/its/pom-deps/verify.groovy
Assertion failed: 

assert new File(basedir, 'target/pom-deps.hpi').exists()
       |        |                               |
       |        |                               false
       |        /home/jenkins/workspace/ools_maven-hpi-plugin_PR-56-AORYQTUGQ4LXFULWYYTKNYE3PDFKEQLVGLEFYDAZMSYHMVBNRHHQ/target/its/pom-deps
       /home/jenkins/workspace/ools_maven-hpi-plugin_PR-56-AORYQTUGQ4LXFULWYYTKNYE3PDFKEQLVGLEFYDAZMSYHMVBNRHHQ/target/its/pom-deps/target/pom-deps.hpi
```

src/it/pom-deps/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+
+    <name>PluginWithPomDependencues</name>
+
+    <dependencies>
+      <dependency>
+        <!-- https://issues.jenkins-ci.org/browse/JENKINS-24374 -->
+        <groupId>org.apache.maven.scm</groupId>
+        <artifactId>maven-scm-providers-standard</artifactId>
+        <version>1.4</version>
+        <type>pom</type>
```
jtnord: it includes the dependencies of it IIUC[1]. (import scope uses dependencyManagement section not the dependencies)

[1] may be a bug rather than a feature @stephenc?

jglick: As written above, I think this is simply a mistake in the upstream POM, trivially worked around in the plugin POM.

jglick:
Forgotten?

PR #57, opened 2017-03-28
Publish plugin manifest as an artifact

ydubreuil:
This pull request is a proposal to publish the manifest included inside the HPI file of a plugin as a separate artifact.

The rationale behind this is to simplify the building of new tools dealing with plugins. It is for example quite difficult to write an off-line plugin bundler right now, as there is no way to list the dependencies of a plugin easily:

* update center JSON file only contains the dependencies of the latest version of each plugin. Dependencies of previous versions are not available
* even if dependencies are available in the POM file, this requires Maven to extract the dependencies graph. Tools we want to build are not always running as a Maven plugin

The [update center updater code](https://github.com/jenkins-infra/backend-update-center2) relies
on the content of this manifest to work, hence it downloads all plugins to be able to generate the JSON update center file (correct me if I'm wrong here, but that's what I understood from the codebase).

There's currently multiple situation where we are working around this situation:

* the official Jenkins docker image offers a [script to bundle plugins](https://github.com/jenkinsci/docker/blob/master/install-plugins.sh) during the creation of an image but people are complaining their builds are not reproducible (which is the reason why BlueOcean image is not using it). The problem is that this script can only download latest plugin version of plugin dependencies.
* BlueOcean is publishing a Docker image with BO plugins bundled and uses [Maven with a fake POM to bundle them](https://github.com/jenkinsci/blueocean-plugin/blob/master/docker/official/pom.xml#L45) instead of the script available in the Docker image

With regard to the Docker script, I already discussed the issue and proposed to use Maven with a fake POM but [this was not seen as a good solution](https://github.com/jenkinsci/docker/issues/327#issuecomment-240022912). It was for sure not a clean one.

At CloudBees, we have internal tooling which has to deal with the situation and having those manifests available would be really helpful. I'm sure there are other examples in the community where having these manifests would be very helpful too. Any tool dealing with plugins needs that information and there's currently no easy way to extract it.

I know it's going to take a long time to have all the manifests published for all plugins but this problem has to be addressed at some point. I wonder if it could be possible to retroactively publish them, btw.

I'm happy to hear what you think about this proposal.

@reviewbybees cc @i386 @aheritier @alecharp @daniel-beck @stephenc

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
@@ -17,7 +17,10 @@ package org.jenkinsci.maven.plugins.hpi;
  */
 
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.PrintWriter;
 import java.util.List;
 import org.apache.commons.lang.StringUtils;
 import org.apache.maven.archiver.MavenArchiveConfiguration;
@@ -159,6 +162,13 @@ public class HpiMojo extends AbstractHpiMojo {
 
         // generate war file
         buildExplodedWebapp(getWebappDirectory(),jarFile);
+
+        File hpiMfFile = getOutputFile(".hpi.mf");
+        getLog().info("Archiving hpi manifest" + hpiMfFile.getAbsolutePath());
+        try (PrintWriter printWriter = new PrintWriter(new OutputStreamWriter(new FileOutputStream(hpiMfFile), "UTF-8"))) {
+            manifest.write(printWriter);
+        }
+        projectHelper.attachArtifact(project, "hpi.mf", null, hpiMfFile);
 
         File hpiFile = getOutputFile(".hpi");
         getLog().info("Generating hpi " + hpiFile.getAbsolutePath());


src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -160,6 +163,13 @@ private void performPackaging()
         // generate war file
         buildExplodedWebapp(getWebappDirectory(),jarFile);
 
+        File hpiMfFile = getOutputFile(".hpi.mf");
```
aheritier: hpi or jpi ?? :-P

daniel-beck:
> Any tool dealing with plugins needs that information and there's currently no easy way to extract it.
> 
> I know it's going to take a long time to have all the manifests published for all plugins but this problem has to be addressed at some point.

For this reason I don't think this is a particularly useful change (or I'm not understanding the short to medium term benefits), as it appears to require all plugins of interest to publish these manifests for this change to make a difference in any tooling using plugin artifacts – even actively maintained plugins don't use recent parent POMs, so I see no hope for the "long tail" of mostly unmaintained plugins. And while this manifest file isn't commonly available, any tooling supporting it may actually slow down further, as an additional artifact will need to be checked for.

ydubreuil:
I agree with you @daniel-beck that there's no immediate short term benefit to this change unless we are able to retroactively publish these missing manifest artifacts. If we can't do that, we'll have to wait maybe a year to start seeing the benefits of this change. But if we don't do it, in a year, we will be at the same point.

> And while this manifest file isn't commonly available, any tooling supporting it may actually slow down further, as an additional artifact will need to be checked for.

Sure, tools will need to implement a fallback to download the HPI file in case the manifest is not directly available.

ndeloof:
👍  this would be useful as well to re-implement https://github.com/jenkinsci/docker/blob/master/install-plugins.sh#L83

jglick:
And Gradle-based plugins would have to follow suit, etc.

Why do you need this anyway? The `*.hpi` artifact already includes the manifest. You need merely extract one entry from a ZIP file. Yes the plugin needs to be downloaded somehow (such as to the local Maven repository), but to do anything useful with the plugin you would need to do that anyway.

ydubreuil:
Having to download the plugin is exactly what I'd like to avoid. Right now, there's no way to index all plugins to extract dependencies relationship to create a custom update center for example, without downloading them all, which takes hours and it'll be worse and worse.

jtnord:
> Having to download the plugin is exactly what I'd like to avoid

Nexus allows you to do that - does Artifactory not allow you to do something similar?

> Right now, there's no way to index all plugins to extract dependencies relationship to create a custom update center for example, without downloading them all, which takes hours and it'll be worse and worse.

All that information is included in the POM and that is available... if non Maven build tools are pushing incorrect POMs for plugins then shame on them  :-)

ydubreuil:
> Nexus allows you to do that - does Artifactory not allow you to do something similar?

You mean, extracting a file from an artifact? If that's possible, I'm interested in :)

> All that information is included in the POM and that is available... if non Maven build tools are pushing incorrect POMs for plugins then shame on them :-)

Right, it's included in the POM, but you need Maven to parse it. I don't really want to re-implement the Maven logic in a custom tool.

jglick:
> there's no way to index all plugins to extract dependencies relationship to create a custom update center for example, without downloading them all

You can cache them in the local repository and then you are only downloading new plugin updates.

daniel-beck:
> Right now, there's no way to index all plugins to extract dependencies relationship to create a custom update center for example, without downloading them all, which takes hours and it'll be worse and worse.

This PR is not the way to resolve this though as it won't actually accomplish your goal IMO. It will simply take too long for a reasonable fraction of plugins to pick up the change. So this won't come close to eliminating downloading & extraction, which means you still need a local Maven repo, which means you've gained nothing except perhaps a few GB disk space saving.

daniel-beck:
@ydubreuil 

> You mean, extracting a file from an artifact? If that's possible, I'm interested in :)

https://www.jfrog.com/confluence/display/RTF/Artifactory+REST+API#ArtifactoryRESTAPI-ArchiveEntryDownload ?

ydubreuil:
Excellent @daniel-beck ! I'm gonna try that

ydubreuil:
I confirm that using the Artifactory REST API to retrieve MANIFEST.MF works on repo.jenkins-ci.org. 

I can't find a similar feature in Nexus though.

oleg-nenashev:
@daniel-beck @ydubreuil Is anyone still interested in getting it over the line?

olamy:
no one interested by this anymore

PR #58, opened 2017-04-11
Escape > as well as < in HTML for compatibility with JDK 8

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
@@ -161,7 +161,7 @@ public class TagLibInterfaceGeneratorMojo extends AbstractMojo {
 
                         JDocComment javadoc = m.javadoc();
                         if (doc!=null)
-                            javadoc.append(doc.getText().replace("&","&amp;").replace("<","&lt;"));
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
```diff
@@ -161,7 +161,7 @@ public boolean accept(File dir, String name) {
 
                         JDocComment javadoc = m.javadoc();
                         if (doc!=null)
-                            javadoc.append(doc.getText().replace("&","&amp;").replace("<","&lt;"));
+                            javadoc.append(doc.getText().replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;"));
```
KostyaSha: No good escaper in EscapeUtils?

jglick: Maybe, I did not look. This suffices.

jglick:
Feel free to file a more elaborate change that would cover more (hypothetical) cases. I am making the minimal change which allows core Javadoc generation to pass.

KostyaSha:
AFAIK javadoc is html, so why not escape it with html escaper?

jglick:
Well the test is that the core build passes!

PR #60, opened 2017-04-30
[JENKINS-43950] Resolve dependencies if jenkins war file does not exist

carlossg:
[JENKINS-43950](https://issues.jenkins-ci.org/browse/JENKINS-43950)

Fail fast instead of later on


carlossg:
Added the javadoc

jglick:
Could you `git merge master` please so we can get a new test run?

oleg-nenashev:
@carlossg ping

carlossg:
rebased

jglick:
This seems to be solving issues similar to #63. I am not sure what the correct approach is here. Probably deserves deeper investigation into the Maven internals.

jglick:
Check whether this is still reproducible after #63.

carlossg:
Fixed with #63

PR #63, opened 2017-05-23
Fix some exceptions breaking `mvn clean hpi:run`

jglick:
Observed in `workflow-job-plugin@master`, using 2.29 parent, that `mvn clean hpi:run` (with no `package` in between, for example) would fail with

```
java.lang.NullPointerException
	at org.jenkinsci.maven.plugins.hpi.MavenArtifact.getActualArtifactId(MavenArtifact.java:146)
	at org.jenkinsci.maven.plugins.hpi.RunMojo.execute(RunMojo.java:333)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
```

Fixed that and then got another exception:

```
java.lang.NullPointerException
	at org.jenkinsci.maven.plugins.hpi.RunMojo.configureWebApplication(RunMojo.java:460)
	at org.jenkinsci.maven.plugins.hpi.AbstractJettyMojo.startJetty(AbstractJettyMojo.java:502)
```

The second fix makes it work. Perhaps there is a simpler solution.

@reviewbybees


jglick:
```
[ERROR] Parser Exception: META-INF/maven/site.vm
org.apache.velocity.runtime.parser.ParseException: Encountered "source" at line 1039, column 52.
Was expecting one of:
    "," ...
    ")" ...
    <WHITESPACE> ...
    
	at org.apache.velocity.runtime.parser.Parser.generateParseException(Parser.java:3679)
	at …
```

Not sure what that is about, checking…

jglick:
Happens in `master` too, will check it separately.

jglick:
@reviewbybees done

KostyaSha:
What introduced this issue?

jglick:
@KostyaSha see #64.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -81,7 +83,7 @@
  * 
  * @author Kohsuke Kawaguchi
  */
-@Mojo(name="run")
+@Mojo(name="run", requiresDependencyResolution=ResolutionScope.TEST)
```
jglick: Hmm, however there is also the `dependencyResolution` parameter, used at the start of `execute`; and `resolveDependencies("test")` in `getJenkinsWarArtifact`. Yet these do not work. @stephenc any clue what is going on here?

stephenc: I think *something* is attaching artifacts to the reactor and then they get wiped by clean and the re-attach is appending to the list, not replacing... does it happen with Maven 3.2.x?

jglick: IIRC I was testing with 3.3.9 at the time.

jglick:
AFAIK the change is helpful and harmless, but withdrawing from consideration until I have time to dig more deeply into why the problem is happening and to check if there is a better solution.

jglick:
I guess I do not actually have time to dig more deeply, and this fix helps people, so :man_shrugging:

carlossg:
This fixes [JENKINS-43950](https://issues.jenkins-ci.org/browse/JENKINS-43950)

PR #62, opened 2017-05-06
upgrade to Jetty 9.4.5.v20170502

olamy:
Note the version bump to 2.0 as it's now Java 8 required.
I upgraded few versions in the pom as well.

PR #64, opened 2017-05-23
Fixing site generation

jglick:
#62 broke the build. I do not know why, exactly; something to do with a Velocity upgrade. I cannot find the file mentioned anywhere. Anyway, should not have been merged until that was corrected, so here it is belatedly so that other PRs do not get marked as failures.

@reviewbybees


jglick:
@reviewbybees done

PR #65, opened 2017-05-24
[JENKINS-24064] Replace war-for-test classifier with executable-war type

jglick:
[JENKINS-24064](https://issues.jenkins-ci.org/browse/JENKINS-24064)

Should allow us to avoid deploying two copies of `jenkins.war`, while simplifying how `WarExploder` finds the WAR file to use.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -724,7 +724,7 @@ public class RunMojo extends AbstractJettyMojo {
             if (jenkinsWarId!=null)
                 match = (a.getGroupId()+':'+a.getArtifactId()).equals(jenkinsWarId);
             else
-                match = (a.getArtifactId().equals("jenkins-war") || a.getArtifactId().equals("hudson-war")) && a.getType().equals("war");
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -724,7 +724,7 @@ protected Artifact getJenkinsWarArtifact() throws MojoExecutionException {
             if (jenkinsWarId!=null)
                 match = (a.getGroupId()+':'+a.getArtifactId()).equals(jenkinsWarId);
             else
-                match = (a.getArtifactId().equals("jenkins-war") || a.getArtifactId().equals("hudson-war")) && a.getType().equals("war");
+                match = (a.getArtifactId().equals("jenkins-war") || a.getArtifactId().equals("hudson-war")) && a.getType().equals("executable-war");
```
stephenc: May want to fallback to finding the old one if this newer match fails

jglick: Yeah, probably could do that, if we wanted to allow the newer plugin to be used in older POMs.

svanoort:
Note that this introduced [JENKINS-45245](https://issues.jenkins-ci.org/browse/JENKINS-45245)

PR #67, opened 2017-06-23
Fixing unchecked/rawtypes warnings in InjectedTest

jglick:
Compilability is verified by existing ITs; I did not bother adding an assertion that there are no compiler warnings in the log since this is pretty trivial.

@reviewbybees


jglick:
Does not really matter (in fact `Map<String, String>` would work AFAICT); just wanted to get rid of the warning.

PR #69, opened 2017-07-24
What's "ea"? Seems arbitrary.

amuniz:
I've found this weird especial version number handle.

Sometimes it is a practice to use short hash strings in versions to identify a release number (because, well, why not? At the end for the day it is as valid as a tag). Today it happened I released an internal artifact with version `1.1.5-ea01b41`. Yes, bad luck, really bad luck (didn't calculate the odds, but I bet they really low).

So `mvn hpi:run` in a project depending on the `1.1.5-ea01b41` ends with `java.lang.NumberFormatException`.

It is not my intention to remove this, just to start a discussion about keeping it or not (and found a PR is the best place to discuss it).

@reviewbybees @jenkinsci/code-reviewers

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/VersionNumber.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/VersionNumber.java
@@ -45,10 +45,6 @@ import java.util.StringTokenizer;
  *
  * <h2>Special tokens</h2>
  * <p>
- * We allow a component to be not just a number, but also "ea", "ea1", "ea2".
- * "ea" is treated as "ea0", and eaN &lt; M for any M &gt; 0.
- *
- * <p>
  * '*' is also allowed as a component, and '*' &gt; M for any M &gt; 0.
  *
  * <p>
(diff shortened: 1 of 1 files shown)


stephenc:
ea == early access

though imho we should just follow the well-known qualifiers that Maven uses: https://github.com/apache/maven/blob/master/maven-artifact/src/main/java/org/apache/maven/artifact/versioning/ComparableVersion.java#L43

oleg-nenashev:
This code came from Jenkins core 8 years ago.
@amuniz Have you checked if there are any usages of this "ea" qualifier?
I rather agree we should get rid of it. Fortunately there is no other qualifiers colliding with hex.

amuniz:
> Have you checked if there are any usages of this "ea" qualifier?

GH does not accept `<` and `>` in searches, but [this](https://github.com/search?q=org%3Ajenkinsci+%22version+ea%22) and [this](https://github.com/search?utf8=%E2%9C%93&q=org%3Ajenkinsci+%22qualifier+ea%22) does not find anything, so no usages on `jenkinsci` at least.

stephenc:
I do not believe that `ea` was ever actually used, but that's what it was for

kzantow:
FYI, `ea` is for "Early Access" in some circles of general naming conventions.

PR #72, opened 2017-08-21
Add Jenkinsfile to create mojo for new plugins

slide:
This adds a simple Jenkinsfile to new plugins created with hpi:create.


hpi-archetype/Jenkinsfile
```diff
@@ -0,0 +1,3 @@
+// This file is used by the Jenkins' project continuous integration infrastructure
+// to build the plugin and verify there are no issues with pull requests
+buildPlugin()
```
oleg-nenashev: line break in the end would be useful

slide:
Is this good to merge now then?

jglick:
@daniel-beck points out that this assumes that the user in fact wants to host the plugin on @jenkinsci, which is not necessarily the case, and the build will fail in other environments. For example [here](https://github.com/jenkinsci/maven-hpi-plugin/blob/137e435d0770f78edbbd2ed54d91809b7e7be5aa/hpi-archetype/pom.xml#L50-L56) we only include _commented-out_ configuration for @jenkinsci hosting. Perhaps this should rather be done as a recommendation within the hosting process for new plugins, such as @slide’s review bot?

slide:
Yeah, very true. I can add as a check in the bot

slide:
I will add a check for a Jenkinsfile to the automated checker.

PR #73, opened 2017-09-02
[JENKINS-41196] config flag for core component override.

kohsuke:
A part of [JENKINS-41196](https://issues.jenkins-ci.org/browse/JENKINS-41196) fix that allows a plugin to be marked as core component override.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -458,16 +458,6 @@ public class RunMojo extends AbstractJettyMojo {
         
         super.configureWebApplication();
         getWebAppConfig().setWar(webAppFile.getCanonicalPath());
-        for (Artifact a : (Set<Artifact>) project.getArtifacts()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -458,16 +458,6 @@ public void configureWebApplication() throws Exception {
         
         super.configureWebApplication();
         getWebAppConfig().setWar(webAppFile.getCanonicalPath());
-        for (Artifact a : (Set<Artifact>) project.getArtifacts()) {
-            if (a.getGroupId().equals("org.jenkins-ci.main") && a.getArtifactId().equals("jenkins-core")) {
-                File coreBasedir = pluginWorkspaceMap.read(a.getId());
-                if (coreBasedir != null) {
-                    String extraCP = new File(coreBasedir, "src/main/resources").toURI() + "," + new File(coreBasedir, "target/classes").toURI();
```
jglick: So IIUC this is removing a fairly important aspect of `hudson-dev:run`. Without this, we would have to restart Jenkins every time we edited any resource.

batmat:
Closing as stalled and conflicted since 1+ year. Will be reopened when/if works on this is rebooted.

PR #76, opened 2017-09-13
[JENKINS-46845] Use the webAppFile parameter

andresrc:
Fixes [JENKINS-46845](https://issues.jenkins-ci.org/browse/JENKINS-46845)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -93,7 +93,7 @@ public class RunMojo extends AbstractJettyMojo {
      * But this parameter allows that to be overwritten.
      * </p>
      */
-    @Parameter
+    @Parameter(property = "webAppFile")
     private File webAppFile;
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -93,7 +93,7 @@
      * But this parameter allows that to be overwritten.
      * </p>
      */
-    @Parameter
+    @Parameter(property = "webAppFile")
```
andresrc: To be able to use it from the command line:

```
mvn hpi:run -DwebAppFile=/path/to/my/custom/jenkins.war
```

batmat: I think using a more qualified name I think `property="hpi.webAppFile` would be slightly better, but not feeling that strongly since anyway the existing code and parameters are not doing this.

armfergom:
🐝

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
andresrc: Yes, I agree, but I've followed the approach of the existing ones.

varyvol:
🐝

PR #77, opened 2017-09-19
[JENKINS-45245] Remove executable-war, switching back to war-for-test

abayer:
[JENKINS-45245](https://issues.jenkins-ci.org/browse/JENKINS-45245)

Specifically rolling back JENKINS-24064 changes.

cc @reviewbybees @jglick @stephenc @kohsuke @daniel-beck

### DIFF
--- a/src/main/resources/META-INF/plexus/components.xml
+++ b/src/main/resources/META-INF/plexus/components.xml
@@ -122,18 +122,5 @@
       </configuration>
     </component>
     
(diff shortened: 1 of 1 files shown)


abayer:
@jglick Again, vehemently disagree. This should be done first, and then undone if/when a non-kludgy solution is found for not having `war-for-test` around and running tests in Intellij.

abayer:
Not needed any more thanks to https://github.com/jenkinsci/jenkins-test-harness/pull/73

PR #80, opened 2018-05-11
[FIXED JENKINS-51264] Add a snapshotPluginVersionOverride configuration option to hpi and hpl goals

stephenc:
See [JENKINS-51264](https://issues.jenkins-ci.org/browse/JENKINS-51264)

### DIFF
--- a/src/it/snapshot-version-override/pom.xml
+++ b/src/it/snapshot-version-override/pom.xml
@@ -1,0 +1,69 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    
+    <modelVersion>4.0.0</modelVersion>
+
+  <parent>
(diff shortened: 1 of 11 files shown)


stephenc:
screen shot 2018-05-11 at 14 44 41

src/it/snapshot-version-override/pom.xml
```diff
@@ -0,0 +1,69 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    
+    <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>1.424</version>
```
batmat: NIT: do we use an outdated parent pom on purpose here?

stephenc: Was a copy of an existing IT, so I kept everything else the same to ensure that this IT is only testing the new property

PR #81, opened 2018-05-21
Fixing RunMojo dependency copy logic

jglick:
Various bugs. For example, `mvn -f artifact-manager-s3 hpi:run` was producing nonsense like

```
[WARNING] will not overwrite …/artifact-manager-s3-plugin/work/plugins/null.jpi with …/.m2/repository/org/jenkins-ci/ui/handlebars/1.1.1/handlebars-1.1.1.hpi because it is newer
```

First of all, this was not a plugin dependency; there was a dependency on `handlebars-1.1.1-core-assets.jar` via `jenkins-war`, not on `handlebars-1.1.1.hpi`. (Similarly, `test`-scoped dependencies on `*-tests.jar` were messing things up.) Second, the mojo was copying into a file `null.jpi`, which wreaked havoc later.

The warning was also inappropriate even in cases where the data was correct—nothing is broken here, just a working home with some plugin updates—and failed to specify the newer version.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/MavenArtifact.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/MavenArtifact.java
@@ -202,6 +202,10 @@ public class MavenArtifact implements Comparable<MavenArtifact> {
             if(!type.equals("jar")) {
                 return type;
             }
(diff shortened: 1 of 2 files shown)


jglick:
IT failures, investigating.

PR #82, opened 2018-05-28
 Simplify creation of manifest entry for JEP-200 for hpi packaging

gladnik:
Allows to add a `Jenkins-ClassFilter-Whitelisted` manifest entry without `maven-jar-plugin` usage.
That is useful, when your pom specifies `hpi` packaging for the plugin.

**Usage:**
Add `<jenkinsClassFilterWhitelisted>true</jenkinsClassFilterWhitelisted>` configuration parameter for the `maven-hpi-plugin`:
```
<plugin>
    <groupId>org.jenkins-ci.tools</groupId>
    <artifactId>maven-hpi-plugin</artifactId>
    <configuration>
        <jenkinsClassFilterWhitelisted>true</jenkinsClassFilterWhitelisted>
    </configuration>
</plugin>
```

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -75,6 +75,7 @@ import org.codehaus.plexus.util.InterpolationFilterReader;
 import org.codehaus.plexus.util.StringUtils;
 
 public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
+
     /**
      * The directory for the generated WAR.
      */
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
@gladnik Why would you need the manifest entry for HPI packaging? All classes in plugins are whitelisted by default

gladnik:
> All classes in plugins are whitelisted by default  

In that case we should close this PR. I'm sorry for your time :(

I did these changes when I tried to fix a plugin. Creation of a resource file for class filter has solved my issue, but I thought that somebody might still need to add Jenkins-ClassFilter-Whitelisted entry for hpi packaging. Seemingly they would not.

oleg-nenashev:
Closing the PR according to the response above. Thanks anyway @gladnik! If you have a Jenkins JIRA ticket for the bug you refer, I would appreciate the link (for https://wiki.jenkins.io/display/JENKINS/Plugins+affected+by+fix+for+JEP-200)

gladnik:
Thank you! There is no ticket, it's our internal plugin.

PR #83, opened 2018-09-07
catch IOException and add useful information to locate problems

ileler:


src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -409,8 +409,13 @@ private VersionNumber versionOfPlugin(File p) throws IOException {
         if (!p.isFile()) {
             return new VersionNumber("0.0");
         }
-        JarFile j = new JarFile(p);
+        JarFile j = null;
```
jglick:
```suggestion
        JarFile j;
        try {
            j = new JarFile(p);
        } catch (IOException x) {
            throw new IOException("not a valid JarFile: " + p, x);
        }
```

so that we do not need null checks.

oleg-nenashev:
Thanks @ileler !

oleg-nenashev:
It was released in `3.0`

PR #86, opened 2018-11-29
parameter to add this plugin to custom war

vtitov:
Corresponding jira issue:
https://issues.jenkins-ci.org/browse/JENKINS-54949

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
@@ -6,14 +6,17 @@ import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.MavenProjectHelper;
-import org.apache.maven.project.artifact.AttachedArtifact;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.taskdefs.Zip;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
```diff
@@ -58,7 +67,18 @@ public void execute() throws MojoExecutionException {
             rezip.addZipfileset(z);
 
             getProject().setArtifacts(resolveDependencies(dependencyResolution));
-            for( MavenArtifact a : getProjectArtifacts() ) {
+
+            Set<MavenArtifact> projectArtifacts = new LinkedHashSet<>(getProjectArtifacts());
+            if(addThisPluginToCustomWar) {
```
oleg-nenashev: Would be nice to ensure that the artifact actually has the `hpi` packaging.

oleg-nenashev:
LGTM, just some defensive code needed.
CC @raul-arabaolaza , because https://github.com/jenkinsci/custom-war-packager has a different default value for this case. I hit it in ci.jenkins.io-runner, and I will follow-up with a pull request

src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
vtitov: added packaging type check in commit [de594d9](https://github.com/jenkinsci/maven-hpi-plugin/pull/86/commits/de594d96b895b696a327dd231866d6159b48f9e0)

oleg-nenashev: thanks!

oleg-nenashev:
@vtitov thanks a lot, will release it today

PR #75, opened 2017-09-10
[JENKINS-20679] Try to declare the minimum Java version in the manifest

daniel-beck:
[JENKINS-20679](https://issues.jenkins-ci.org/browse/JENKINS-20679)

Only tested manually, including with a patched plugins POM to see whether the compiler plugin configuration fallback works. Can rip that out and we only support property or HPI plugin configuration for this.

Downstream PRs are https://github.com/jenkins-infra/backend-update-center2/pull/164 and https://github.com/jenkinsci/jenkins/pull/3016.

Upstream of: https://github.com/jenkinsci/plugin-pom/pull/134

@jenkinsci/code-reviewers

PR #88, opened 2018-12-14
Pass minimumJavaVersion to RunMojo so that hpi:run works again

jglick:
Corrects a regression in #75, noticed in https://github.com/jenkinsci/external-monitor-job-plugin/pull/18.

### DIFF
--- a/src/it/parent-3x/invoker.properties
+++ b/src/it/parent-3x/invoker.properties
@@ -1,3 +1,3 @@
 # install, not verify, because we want to check the artifact as we would be about to deploy it
 # release.skipTests normally set in jenkins-release profile since release:perform would do the tests
-invoker.goals=-Pjenkins-release -Drelease.skipTests=false clean install
+invoker.goals=-Pjenkins-release -Drelease.skipTests=false clean install hpi:run
--- a/pom.xml
+++ b/pom.xml
@@ -360,7 +360,6 @@
                 </goals>
                 <configuration>
                   <streamLogs>false</streamLogs>
-                  <debug>true</debug>
                 </configuration>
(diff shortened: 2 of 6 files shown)


src/it/parent-3x/invoker.properties
```diff
@@ -1,3 +1,3 @@
 # install, not verify, because we want to check the artifact as we would be about to deploy it
 # release.skipTests normally set in jenkins-release profile since release:perform would do the tests
-invoker.goals=-Pjenkins-release -Drelease.skipTests=false clean install
+invoker.goals=-Pjenkins-release -Drelease.skipTests=false clean install hpi:run
```
oleg-nenashev: I am not sure it is a good idea to invoke hpi:run in the test. What is an expected behavior?

jglick:
> the test automation approach does not look to be very reliable

Well the test fails with the observed error when the fix is commented out, and passes with the fix in place. Do you have a better idea?

jglick:
Hmm

```
org.apache.maven.plugin.MojoExecutionException: Error preparing the manifest: /var/jenkins_home/plugins/parent-3x.hpl (No such file or directory)
    at org.jenkinsci.maven.plugins.hpi.HplMojo.execute (HplMojo.java:108)
    at org.jenkinsci.maven.plugins.hpi.RunMojo.generateHpl (RunMojo.java:480)
    at org.jenkinsci.maven.plugins.hpi.RunMojo.execute (RunMojo.java:329)
```

is something I do not get locally.

jglick:
Ah, the test fails when the `JENKINS_HOME` environment variable is set…

PR #91, opened 2019-01-26
[WIP] [JENKINS-28942] Add property to instruct the Core to not add implied dependencies

batmat:
[JENKINS-28942](https://issues.jenkins-ci.org/browse/JENKINS-28942)

* [x] Code adding the property, also to gather agreement on the naming **`noDetachedPluginUntil`**
* [ ] add Maven invoker IT for the new property

After this, there will need also:
* a change against the update-center generation repo so the property is extracted and added to JSON (https://github.com/jenkins-infra/update-center2/)
* a change in Jenkins Core to use that property to filter the implied dependencies depending on the property value (see https://github.com/jenkinsci/jenkins/blob/226f7b4c2bedb14b70f12da90db15574e18364d0/core/src/main/resources/jenkins/split-plugins.txt#L9-L30)

cc @jenkinsci/java11-support ([given I am doing this because this is the right way to handle my main concern about JAXB addition as a detached plugin](https://issues.jenkins-ci.org/browse/JENKINS-55681): that doing so nobody will ever have an actual way to break the dependency against JAXB)

@daniel-beck @jglick who commented on the JIRA.

**While this is still WIP for missing tests, please validate the `noDetachedPluginUntil` name of the attribute so that I have no back and forths to do when doing the associated PR in others repos.**

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
@@ -73,6 +73,15 @@ public abstract class AbstractJenkinsManifestMojo extends AbstractHpiMojo {
     protected String minimumJavaVersion;
 
     /**
+     * Indicates the plugin should not get any implied dependency until the specified (Jenkins core) version included.
+     * <p>Example: if a plugin specifies this with a value of <em>2.60.3</em>, and used bouncycastle API (which got detacted in 2.16) without explictly
+     * declaring the dependency against bouncycastle-api, then the plugin would likely fail at runtime because Jenkins Core Plugin Management will
+     * then <strong>not</strong> add bouncycastle-api as an implicit dependency.</p>
+     */
+    @Parameter(property = "hpi.noDetachedPluginUntil", required = true)
+    protected String noDetachedPluginUntil;
+
+    /**
      * Generates a manifest file to be included in the .hpi file
      */
     protected void generateManifest(MavenArchiveConfiguration archive, File manifestFile) throws MojoExecutionException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -72,6 +72,15 @@
     @Parameter(required = true)
     protected String minimumJavaVersion;
 
+    /**
+     * Indicates the plugin should not get any implied dependency until the specified (Jenkins core) version included.
+     * <p>Example: if a plugin specifies this with a value of <em>2.60.3</em>, and used bouncycastle API (which got detacted in 2.16) without explictly
+     * declaring the dependency against bouncycastle-api, then the plugin would likely fail at runtime because Jenkins Core Plugin Management will
+     * then <strong>not</strong> add bouncycastle-api as an implicit dependency.</p>
+     */
+    @Parameter(property = "hpi.noDetachedPluginUntil", required = true)
```
jglick: Why `required`? This certainly sounds like it should be optional to me.

batmat: The intent is to have it advertised once we propagate it at some point in the plugin-pom, so plugin developers are hinted about that when bumping the parent.

We'd document there something like:

> If you want all implied dependencies, and want to go yolo: just put `1.0` and move on.
> Better: do X and Y do to actually be able to define a sensible value.

And worst worst case: we put 1.0 as default value in `plugin-pom`.

jglick: Uh, having this new manifest attribute be `1.0`, or generally _less than_ `Jenkins-Version`, seems rather confusing to me. I would really recommend the attribute be omitted unless you are actually using this feature. Would suffice to make this parameter optional, and treat null or `""` as unset. I am not even sure this is something we want to advertise in `plugin-pom`; the small number of plugins which need this could define `hpi.noDetachedPluginUntil`, and `buildPlugin.groovy` could detect this and automatically include that version in the testing matrix.

batmat:
Closing for now, I hope to get back to this.

PR #104, opened 2019-03-22
A ProjectBuildingException on some dependency should not be fatal

jglick:
E.g. https://github.com/kubernetes-client/java/pull/334#issuecomment-475674883. Maven already prints a warning about such dependencies when computing the set of artifacts, as does the Enforcer plugin, so this is just preventing later phases from breaking the build.


stephenc:
🤷‍♂️

jglick:
Trying to release, but looks like I need to wait for https://github.com/jenkins-infra/repository-permissions-updater/pull/1074.

runzexia:
Hi all the dependency on kubernetes-client in the jenkins plugin still fails.
https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fkubernetes-cd-plugin/detail/PR-87/12/pipeline

PR #118, opened 2019-06-10
Use Maven release plugin 2.5.3 to enable releases on Windows

oleg-nenashev:
Just a quick patch to get the release flow running


oleg-nenashev:
Upper bounds is a bloody mess in this component, so I prefer not to touch it for now

oleg-nenashev:
@jglick done, https://github.com/jenkinsci/maven-hpi-plugin/pull/118#issuecomment-500467930 answers why I don;t update it

PR #132, opened 2019-08-19
Align versions of Maven and Maven Plugin Tools dependencies

oleg-nenashev:
I would like to update to Maven 3.5 later to keep it aligned with Plugin POM, but such change should at least help with dependency management a bit


jglick:
> update to Maven 3.5 later

Tricky. See #23.

oleg-nenashev:
> Tricky

It is. Anyway, this change will not hurt as is.

PR #138, opened 2019-08-26
Adding LICENSE file to project

kshithijiyer:


jglick:
I generally consider the POM to be authoritative, so no additional file is needed, unless there is some independent reason to create a file of this name?

oleg-nenashev:
GitHub repository metadata for a use-cases FWIW

oleg-nenashev:
@kshithijiyer Hi. Are you still interested in finishing it?

kshithijiyer:
Yes! There are no changes needed as far as I understand right ?

oleg-nenashev:
There are: https://github.com/jenkinsci/maven-hpi-plugin/pull/138#pullrequestreview-286150081

kshithijiyer:
Normally in opensource projects license is stored in either a LICENSE file or a Copying or copyright file. pom.xml isn't a place where people would generally look for if they want to know how the project is open sourced.

oleg-nenashev:
The only blocker is that the pom.xml declares Apache Software License v.2 now @daniel-beck .
At the same  some code pieces are licensed as MIT or not explicitly licensed at all.

I am fine with aligning the licenses to MIT-only or to Apache v.2-only, but we need a decision here.

CC @jglick @stephenc @kohsuke as a most active contributors. The Apache v.2 license was put by @kohsuke , and I suggest sticking to it as a default strategy

oleg-nenashev:
There are only 2 files with explicit MIT specification:

* https://github.com/jenkinsci/maven-hpi-plugin/blob/602d838752ba7f83ed2d2f0b415b2900ba5563fc/src/main/java/hudson/Extension.java
* https://github.com/jenkinsci/maven-hpi-plugin/blob/602d838752ba7f83ed2d2f0b415b2900ba5563fc/src/main/java/jenkins/YesNoMaybe.java

Both components are copied from the Jenkins core AFAICT. I suggest we ignore that for now and proceed with the Apache Software License v.2 in the root

kshithijiyer:
Sounds good to me! The pom.xml thing is already explained in the previous comment:

> Normally in opensource projects license is stored in either a LICENSE file or a Copying or copyright file. pom.xml isn't a place where people would generally look for if they want to know how the project is open sourced.

Are there any other changes needed?

oleg-nenashev:
I implemented the change in https://github.com/jenkinsci/maven-hpi-plugin/pull/147 . Sorry for taking it over @kshithijiyer , it just felt we are running in circles here w.r.t the license terms and the requested changes.

PR #154, opened 2019-10-24
Remove the DEV@cloud reference from the github.io site

oleg-nenashev:
Just because the service is shut down


oleg-nenashev:
We could rework the site later, but whatever

PR #157, opened 2019-11-01
Upgrade Jetty from 9.4.5.v20170502 to 9.4.22.v20191022

jglick:
Supersedes #150. Picking up the sort of no-op servlet security realm that `JenkinsRule` uses. Confirmed that this works with the `build-token-root` plugin with authentication enabled (`initialAdminPassword` style).

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -40,6 +40,7 @@ import org.apache.maven.project.MavenProjectBuilder;
 import org.eclipse.jetty.maven.plugin.JettyWebAppContext;
 import org.eclipse.jetty.maven.plugin.MavenServerConnector;
 import org.eclipse.jetty.security.HashLoginService;
+import org.eclipse.jetty.security.UserStore;
 import org.eclipse.jetty.server.ConnectionFactory;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.HttpConnectionFactory;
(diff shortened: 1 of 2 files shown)


jglick:
> A downstream PR

To do what?

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -503,9 +505,12 @@ public void configureWebApplication() throws Exception {
...
         }
-        // cf. https://wiki.jenkins-ci.org/display/JENKINS/Jetty
-        HashLoginService hashLoginService = (new HashLoginService("Jenkins Realm"));
-        hashLoginService.setConfig(System.getProperty("jetty.home", "work") + "/etc/realm.properties");
+        HashLoginService hashLoginService = (new HashLoginService("default"));
+        UserStore userStore = new UserStore();
+        hashLoginService.setUserStore(userStore);
+        userStore.addUser("alice", new Password("alice"), new String[] {"user", "female"});
+        userStore.addUser("bob", new Password("bob"), new String[] {"user", "male"});
+        userStore.addUser("charlie", new Password("charlie"), new String[] {"user", "male"});
```
jglick: Copied from https://github.com/jenkinsci/jenkins-test-harness/blob/24567f4ee89a2815f436dea84a0f180ff14120af/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L789-L799.

jglick:
> I meant this Build Token Root plugin you referenced

Pointless since what I tested was behavior of `hpi:run` which a CI build does not exercise. Whereas the `parent-3x` IT in this repo _does_ sanity-check it.

basil:
Caused regression (#625)

PR #162, opened 2019-11-25
Fix typo in change log hpi property

timja:
cc @oleg-nenashev

### DIFF
--- a/src/it/JENKINS-45740-metadata/pom.xml
+++ b/src/it/JENKINS-45740-metadata/pom.xml
@@ -15,10 +15,10 @@
         <java.level>8</java.level>
         <hpi-plugin.version>@project.version@</hpi-plugin.version>
         <!--Examples of real use-cases-->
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@timja Released it in https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/3.11 . Since we have only 1 plugin affected, I suggest to not expedite the plugin POM release for now. The version of this plugin can be defined via a property. WDYT?

timja:
> @timja Released it in [`3.11` (release)](https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/3.11) . Since we have only 1 plugin affected, I suggest to not expedite the plugin POM release for now. The version of this plugin can be defined via a property. WDYT?

I think we can just pick it up in the next plugin-pom release (skip the property override).
As long as it's released before it's widely advertised it's fine

PR #169, opened 2020-02-22
Listen on loopback interface by default

daniel-beck:
Local development setups usually don't need to listen on all interfaces, so restrict to `localhost` by default.

Apparently `host` is the Jetty terminology for the listen address, so use that here.
I didn't want to think about IPv4/IPv6, so use `localhost` instead of e.g. `127.0.0.1`. `JenkinsRule` uses `localhost` as well.

Restore previous behavior by setting `-Dhost=0.0.0.0` on the command line.

Corresponding PR in core at https://github.com/jenkinsci/jenkins/pull/4515.


oleg-nenashev:
https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/maven-hpi-plugin-3.12

PR #172, opened 2020-03-05
[JENKINS-58771] Avoid more cases of bundling libs incorrectly

Vlatombe:
Expands https://github.com/jenkinsci/maven-hpi-plugin/pull/140

Sometimes some dependencies gets pulled through test or provided scope
trails, because in fact they are being pulled in compile scope through a plugin with a longer trail.

Inspired by https://github.com/jenkinsci/kubernetes-plugin/pull/729

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -496,23 +496,27 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
         List<File> dependentWarDirectories = new ArrayList<File>();
 
         // List up IDs of Jenkins plugin dependencies
-        Set<String> jenkinsPlugins = new HashSet<String>();
+        Set<String> jenkinsPlugins = new HashSet<>();
+        Set<String> excludedArtifacts = new HashSet<>();
         for (MavenArtifact artifact : Sets.union(artifacts, dependencyArtifacts)) {
             if (artifact.isPluginBestEffort(getLog()))
                 jenkinsPlugins.add(artifact.getId());
             // Exclude dependency if it comes from test or provided trail.
             // Most likely a plugin transitive dependency but the trail through (test,provided) dep is shorter
             if (artifact.hasScope("test", "provided"))
-                jenkinsPlugins.add(artifact.getId());
+                excludedArtifacts.add(artifact.getId());
         }
 
         OUTER:
         for (MavenArtifact artifact : artifacts) {
-            if(jenkinsPlugins.contains(artifact.getId()))
+            if (jenkinsPlugins.contains(artifact.getId()))
                 continue;   // plugin dependency need not be WEB-INF/lib
-            if(artifact.getDependencyTrail().size() >= 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
-                continue;   // no need to have transitive dependencies through plugins in WEB-INF/lib.
-
+            if (artifact.getDependencyTrail().size() >= 1) {
+                if (jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
+                    continue; // no need to have transitive dependencies through plugins in WEB-INF/lib.
+                if (excludedArtifacts.contains(artifact.getDependencyTrail().get(1)))
+                    continue; // exclude artifacts resolved through a test or provided scope
+            }
             // if the dependency goes through jenkins core, we don't need to bundle it in the war
             // because jenkins-core comes in the <provided> scope, I think this is a bug in Maven that it puts such
             // dependencies into the artifact list.
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -503,19 +503,23 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
                 jenkinsPlugins.add(artifact.getId());
             // Exclude dependency if it comes from test or provided trail.
             // Most likely a plugin transitive dependency but the trail through (test,provided) dep is shorter
-            if (artifact.hasScope("test", "provided"))
+            if (artifact.hasScope("test", "provided")) {
                 excludedArtifacts.add(artifact.getId());
+            }
         }
 
         OUTER:
         for (MavenArtifact artifact : artifacts) {
-            if (jenkinsPlugins.contains(artifact.getId()))
+            if (jenkinsPlugins.contains(artifact.getId())) {
                 continue;   // plugin dependency need not be WEB-INF/lib
+            }
             if (artifact.getDependencyTrail().size() >= 1) {
-                if (jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
+                if (jenkinsPlugins.contains(artifact.getDependencyTrail().get(1))) {
                     continue; // no need to have transitive dependencies through plugins in WEB-INF/lib.
-                if (excludedArtifacts.contains(artifact.getDependencyTrail().get(1)))
+                }
+                if (excludedArtifacts.contains(artifact.getDependencyTrail().get(1))) {
                     continue; // exclude artifacts resolved through a test or provided scope
+                }
             }
             // if the dependency goes through jenkins core, we don't need to bundle it in the war
             // because jenkins-core comes in the <provided> scope, I think this is a bug in Maven that it puts such
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -500,6 +500,10 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
         for (MavenArtifact artifact : Sets.union(artifacts, dependencyArtifacts)) {
             if (artifact.isPluginBestEffort(getLog()))
                 jenkinsPlugins.add(artifact.getId());
+            // Exclude dependency if it comes from test or provided trail.
+            // Most likely a plugin transitive dependency but the trail through (test,provided) dep is shorter
+            if (artifact.hasScope("test", "provided"))
+                jenkinsPlugins.add(artifact.getId());
         }
 
         OUTER:
--- a/src/it/JENKINS-58771-packaged-plugins-2/invoker.properties
+++ b/src/it/JENKINS-58771-packaged-plugins-2/invoker.properties
@@ -1,0 +1,1 @@
+invoker.goals=-Dstyle.color=always -ntp clean install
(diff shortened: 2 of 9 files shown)


jtnord:
also seeing cases of

```
[WARNING] Bundling transitive dependency jsr305-3.0.2.jar (via spotbugs-annotations)
[INFO] Bundling direct dependency rate-limited-logger-2.0.0.jar
[WARNING] Bundling transitive dependency eddsa-0.3.0.jar (via jenkins-war)
[WARNING] Bundling transitive dependency slf4j-api-1.7.26.jar (via rate-limited-logger)
```

I only see `eddsa` as test scope from jenkins-war so looks like this will also help.

```
[INFO] +- org.jenkins-ci.main:jenkins-war:executable-war:2.222.1-cb-1:test
[INFO] |  +- org.jenkins-ci.modules:ssh-cli-auth:jar:1.7:test
[INFO] |  +- org.jenkins-ci.modules:slave-installer:jar:1.7:test
[INFO] |  +- org.jenkins-ci.modules:windows-slave-installer:jar:1.12:test
[INFO] |  +- org.jenkins-ci.modules:launchd-slave-installer:jar:1.2:test
[INFO] |  +- org.jenkins-ci.modules:upstart-slave-installer:jar:1.1:test
[INFO] |  +- org.jenkins-ci.modules:systemd-slave-installer:jar:1.1:test
[INFO] |  +- org.jenkins-ci.modules:sshd:jar:2.6:test
[INFO] |  |  +- org.apache.sshd:sshd-core:jar:1.7.0:test
[INFO] |  |  \- net.i2p.crypto:eddsa:jar:0.3.0:test
```

jtnord:
This PR fixes the above issues for me.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -500,6 +500,10 @@ public void buildWebapp(MavenProject project, File webappDirectory)
         for (MavenArtifact artifact : Sets.union(artifacts, dependencyArtifacts)) {
             if (artifact.isPluginBestEffort(getLog()))
                 jenkinsPlugins.add(artifact.getId());
+            // Exclude dependency if it comes from test or provided trail.
+            // Most likely a plugin transitive dependency but the trail through (test,provided) dep is shorter
+            if (artifact.hasScope("test", "provided"))
+                jenkinsPlugins.add(artifact.getId());
```
jglick: Can you explain what is going on here? `jenkinsPlugins` is supposed to be a list of…well, Jenkins plugins. Which `isPluginBestEffort` is supposed to be detecting. So either that method is broken in some cases (why?); or what you are collecting in this set is not in fact a list of Jenkins plugins but just a list of stuff you want to exclude from processing below, in which case it would be clearer to write something like

```diff
diff --git src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
index 431052b..d70f0ab 100644
--- src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
+++ src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
@@ -508,6 +508,9 @@ public abstract class AbstractHpiMojo extends AbstractJenkinsMojo {
                 continue;   // plugin dependency need not be WEB-INF/lib
             if(artifact.getDependencyTrail().size() >= 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
                 continue;   // no need to have transitive dependencies through plugins in WEB-INF/lib.
+            if (artifact.hasScope("test", "provided")) {
+                continue; // TODO some explanation
+            }
 
             // if the dependency goes through jenkins core, we don't need to bundle it in the war
             // because jenkins-core comes in the <provided> scope, I think this is a bug in Maven that it puts such
```

jtnord:
> also seeing cases of
> 
> ```
> [WARNING] Bundling transitive dependency jsr305-3.0.2.jar (via spotbugs-annotations)
> [INFO] Bundling direct dependency rate-limited-logger-2.0.0.jar
> [WARNING] Bundling transitive dependency eddsa-0.3.0.jar (via jenkins-war)
> [WARNING] Bundling transitive dependency slf4j-api-1.7.26.jar (via rate-limited-logger)
> ```
> 
> I only see `eddsa` as test scope from jenkins-war so looks like this will also help.
> 
> ```
> [INFO] +- org.jenkins-ci.main:jenkins-war:executable-war:2.222.1-cb-1:test
> [INFO] |  +- org.jenkins-ci.modules:ssh-cli-auth:jar:1.7:test
> [INFO] |  +- org.jenkins-ci.modules:slave-installer:jar:1.7:test
> [INFO] |  +- org.jenkins-ci.modules:windows-slave-installer:jar:1.12:test
> [INFO] |  +- org.jenkins-ci.modules:launchd-slave-installer:jar:1.2:test
> [INFO] |  +- org.jenkins-ci.modules:upstart-slave-installer:jar:1.1:test
> [INFO] |  +- org.jenkins-ci.modules:systemd-slave-installer:jar:1.1:test
> [INFO] |  +- org.jenkins-ci.modules:sshd:jar:2.6:test
> [INFO] |  |  +- org.apache.sshd:sshd-core:jar:1.7.0:test
> [INFO] |  |  \- net.i2p.crypto:eddsa:jar:0.3.0:test
> ```
> 
> slf4j is being promoted via an `optional` dependency
> 
> ```
> [INFO] com.cloudbees.operations-center.server:operations-center-server:hpi:2.204.2.112-SNAPSHOT
> [INFO] +- org.jenkins-ci.plugins:support-core:jar:2.64:compile (optional)
> [INFO] |  +- net.sf.uadetector:uadetector-resources:jar:2013.09:compile (optional)
> [INFO] |  |  +- net.sf.uadetector:uadetector-core:jar:0.9.9:compile (optional)
> [INFO] |  |  |  \- net.sf.qualitycheck:quality-check:jar:1.3:compile (optional)
> [INFO] |  |  \- org.slf4j:slf4j-api:jar:1.7.26:compile
> ```

eddsa..

image

promoted to compile due to dependency on maven plugin which depends on ssh-credentials which depends on...  whilst correct at the maven level, maven-plugin is a HPI and should break the cycle.

image

atanasenko:
I found usage of `getDependencyTrail()` to be flawed to begin with. It is always the shortest path to dependency and the reason why maven included that particular dependency. 
Logic behind jenkins/hpi conflicts with the way maven performs resolution, shorter trail of, for example, `dep1 -> commons-exec` will always win over `plugindep` -> dep2 -> commons-exec` and will bundle it (irregardless of version differences), though it will never be used unless you mask classes from it.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
amuniz: > jenkinsPlugins is supposed to be a list of…well, Jenkins plugins

Well, it's not. It contains all dependencies, regardless their type. Was badly named since the begging. Anyways, I think @Vlatombe has already put it another way in new commits.

Vlatombe: I separated into a different set for clarity.

Vlatombe:
> I found usage of getDependencyTrail() to be flawed to begin with. It is always the shortest path to dependency and the reason why maven included that particular dependency.

Indeed, not using a specific type (`hpi` or `jpi`) when pulling jenkins plugins dependencies is making it impossible to rely on Maven to do the right thing.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -496,19 +496,27 @@ public void buildWebapp(MavenProject project, File webappDirectory)
...
-        Set<String> jenkinsPlugins = new HashSet<String>();
+        Set<String> jenkinsPlugins = new HashSet<>();
+        Set<String> excludedArtifacts = new HashSet<>();
         for (MavenArtifact artifact : Sets.union(artifacts, dependencyArtifacts)) {
             if (artifact.isPluginBestEffort(getLog()))
                 jenkinsPlugins.add(artifact.getId());
+            // Exclude dependency if it comes from test or provided trail.
+            // Most likely a plugin transitive dependency but the trail through (test,provided) dep is shorter
+            if (artifact.hasScope("test", "provided"))
+                excludedArtifacts.add(artifact.getId());
```
jglick:
```suggestion
            if (artifact.hasScope("test", "provided")) {
                excludedArtifacts.add(artifact.getId());
            }
```

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -496,19 +496,27 @@ public void buildWebapp(MavenProject project, File webappDirectory)
...
         }
 
         OUTER:
         for (MavenArtifact artifact : artifacts) {
-            if(jenkinsPlugins.contains(artifact.getId()))
+            if (jenkinsPlugins.contains(artifact.getId()))
                 continue;   // plugin dependency need not be WEB-INF/lib
```
jglick:
```suggestion
            if (jenkinsPlugins.contains(artifact.getId())) {
                continue;   // plugin dependency need not be WEB-INF/lib
            }
```

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -496,19 +496,27 @@ public void buildWebapp(MavenProject project, File webappDirectory)
...
-            if(artifact.getDependencyTrail().size() >= 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
-                continue;   // no need to have transitive dependencies through plugins in WEB-INF/lib.
-
+            if (artifact.getDependencyTrail().size() >= 1) {
+                if (jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
+                    continue; // no need to have transitive dependencies through plugins in WEB-INF/lib.
+                if (excludedArtifacts.contains(artifact.getDependencyTrail().get(1)))
+                    continue; // exclude artifacts resolved through a test or provided scope
```
jglick:
```suggestion
                if (jenkinsPlugins.contains(artifact.getDependencyTrail().get(1))) {
                    continue; // no need to have transitive dependencies through plugins in WEB-INF/lib.
                }
                if (excludedArtifacts.contains(artifact.getDependencyTrail().get(1))) {
                    continue; // exclude artifacts resolved through a test or provided scope
                }
```

jglick:
`JENKINS-58771-packaged-plugins-2` failed

Vlatombe:
Re-running build

jetersen:
I seen a few more cases of odd bundling when it comes to plugin pom v4 and Jenkins Core bom.
This is using plugin pom v4.2 which uses hpi v3.14.

seems Guava, commons-io, commons-lang are chosen to be bundled in transitive dependencies even though Jenkins core provide these versions.

I will link to PR for artifactory plugin that I will land it shortly

Vlatombe:
@jetersen I looked at the PR you referenced and hpi plugin is behaving as expected. If you depend on a jar that pulls transitive dependencies that may be provided through other means in the Jenkins ecosystem, you have to exclude these transitive dependencies, unless you have a very good reason to include them (like pulling a more recent version).

jetersen:
I added jackson-api plugin expecting those would not be transitively bundled from both buildinfo-api and buildinfo-client

PR #178, opened 2020-04-21
Offer a browse URL from hpi:run including the likes of nip.io if desired

jglick:
When you `hpi:run` from a lot of plugins, each with their own `work` directory, it is annoying to always be using the same `http://localhost:8080/jenkins/` URL: your address bar will get all mixed up, cookies will get crossed over, saved `initialAdminPassword`s will be confused, etc. It is nicer to browse to e.g. `http://build-token-root.127.0.0.1.nip.io:8080/jenkins/` which corresponds to just one `work` directory. This patch just makes it a bit easier to get such a URL.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -162,6 +162,14 @@ public class RunMojo extends AbstractJettyMojo {
     protected String defaultHost;
 
     /**
+     * Optional wildcard DNS domain to help set a distinct Jenkins root URL from every plugin.
+     * Just prints a URL you ought to set.
+     * Recommended: {@code nip.io}
+     */
+    @Parameter(property = "wildcardDNS")
+    protected String wildcardDNS;
+
+    /**
      * If true, the context will be restarted after a line feed on
      * the input console. Disabled by default.
      */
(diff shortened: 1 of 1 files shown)


daniel-beck:
Does this make assumptions about services _like_ nip.io that are only valid _for_ nip.io? I am unfamiliar with them, but wonder whether the pattern used here is universal.

MarkEWaite:
> Does this make assumptions about services _like_ nip.io that are only valid _for_ nip.io? I am unfamiliar with them, but wonder whether the pattern used here is universal.

The xip.io service seems to provide the same type of response as nip.io for this case.  For example:

```
mwaite@mark-pc2:~/git/jenkins/jenkins.io$ dig 127.0.0.1.xip.io
;; ANSWER SECTION:
127.0.0.1.xip.io.       300     IN      A       127.0.0.1
```

```
mwaite@mark-pc2:~/git/jenkins/jenkins.io$ dig 127.0.0.1.nip.io
;; ANSWER SECTION:
127.0.0.1.nip.io.       432000  IN      A       127.0.0.1
```

The assumption that the local IP address prefixed to the DNS resolver basename seems to apply to both those service providers.

jglick:
Right, there are other suffixes that work with this pattern.

jglick:
Minor possible enhancement: if `work/jenkins.model.JenkinsLocationConfiguration.xml` already exists, display the contents of `jenkinsUrl`, in case the user has deliberately picked something different like using ngrok.

PR #191, opened 2020-08-12
Add `validate-hpi` which checks core version for dependencies before packaging

Vlatombe:
Adds check that the required core version is higher or equal to the core version required by plugin dependencies.

This is indirectly done by the injected tests already but it requires to run the tests and it is much longer.


src/it/check-core-version/pom.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>4.6</version>
+    <relativePath />
+  </parent>
```
timja, line 10: shouldn't this test have the jenkins.version set so that when parent bumps the parent pom here and in parent pom the test won't break?

timja (commented):
would be nice to also see a test that show this doesn't break exact version matches

e.g. 2.222.4 is fine in this case.

timja (approved):
Thanks

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
```diff
@@ -1,24 +1,33 @@
...
+public class ValidateMojo extends AbstractHpiMojo {
     public void execute() throws MojoExecutionException, MojoFailureException {
-        if (!isMustangOrAbove())
+        if (!isMustangOrAbove()) {
             throw new MojoExecutionException("JDK6 or later is necessary to build a Jenkins plugin");
-
-        if (new VersionNumber(findJenkinsVersion()).compareTo(new VersionNumber("1.419.99"))<=0)
+        }
+        if (new VersionNumber(findJenkinsVersion()).compareTo(new VersionNumber("1.419.99"))<=0) {
             throw new MojoExecutionException("This version of maven-hpi-plugin requires Jenkins 1.420 or later");
```
oleg-nenashev, line 28: needs to be updated in a separate pull request. We have a decision to drop support of pre-Java 8 Jenkins core versions from development tools, and IMHO it makes sense to apply it here. Can be done in a follow-up PR

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
```diff
@@ -32,4 +41,42 @@ private static boolean isMustangOrAbove() {
...
         }
     }
+
+    private void validateCoreVersion() throws MojoExecutionException {
+        try {
+            VersionNumber coreVersion = new VersionNumber(findJenkinsVersion());
+            MavenArtifact maxCoreVersionArtifact = null;
+            VersionNumber maxCoreVersion = new VersionNumber("0");
+
+            for (MavenArtifact artifact : Sets.union(getProjectArtfacts(), getDirectDependencyArtfacts())) {
```
oleg-nenashev, line 51: Not a blocker: Note that it does not take transitive dependencies into account. All plugins in the dependency tree will have to adopt the new Maven HPI Plugin version to prevent from errors in transitive deps. Related to #195 which also needs a full plugin dependency tree resolution.

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
```diff
@@ -1,24 +1,33 @@
...
 
 /**
  * Make sure that we are running in the right environment.
  *
  * @author Kohsuke Kawaguchi
  */
-@Mojo(name = "validate", defaultPhase = LifecyclePhase.VALIDATE)
-public class ValidateMojo extends AbstractJenkinsMojo {
+@Mojo(name = "validate", defaultPhase = LifecyclePhase.VALIDATE, requiresDependencyResolution = ResolutionScope.RUNTIME)
+public class ValidateMojo extends AbstractHpiMojo {
```
oleg-nenashev, line 22: It will prevent from using the mojo outside the HPI packaging use-cases. I would recommend to retain compatibility and to implement a new mojo instead.

oleg-nenashev (changes requested):
Currently `ValidateMojo` might be used for non-HPI packaging. With this change HPI packaging will be required IIUC the implementation. Please correct me if I am wrong. If not, I suggest creating a new `validate-hpi` mojo for the new checks.

jglick:
Note that you might expect this to be done by `RequireUpperBoundsDeps` but you would be wrong, since `jenkins-core` is a `provided` dep and thus not transitive.

oleg-nenashev (approved):
LGTM. There could be a more fancy documentation, but it should be good enough for adoption in parent POMs.

oleg-nenashev:
https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/maven-hpi-plugin-3.16 is out @Vlatombe . I'd guess we need to enable the new mojo by default in Jenkins Plugin POM.

Vlatombe:
@oleg-nenashev Should be automatic since I added it to the `hpi` packaging https://github.com/jenkinsci/maven-hpi-plugin/pull/191/files#diff-c9fbb41a606335288946706b0925b227R14

oleg-nenashev:
right, let's see. An integration test would be useful in Plugin POM, just in case

jglick:
> An integration test would be useful in Plugin POM

Perhaps, though as noted in https://github.com/jenkinsci/maven-hpi-plugin/pull/191#issue-466738726, a misconfigured plugin build will already fail—it will just take a bit longer.

FWIW I keep a custom shortcut in NetBeans to help test `pom.xml` edits:

```bash
mvn clean test -Dtest=InjectedTest
```

PR #208, opened 2020-12-10
[JENKINS-64408] support targetPath property for webResources

sephiroth-j:
- it works for the packaging phase out of the box
- for use as part of `hpi:run`, the goals `compile` and `hpi:hpi` must be executed before to ensure that `webappDirectory` exists. otherwise `warSourceDirectory` will be used as before.
**example:** `mvn -DskipTests=true -Dport=8081 clean compile hpi:hpi hpi:run`

Please see [JENKINS-64408](https://issues.jenkins.io/browse/JENKINS-64408) for the background and motivation of this.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
@@ -2,7 +2,6 @@ package org.jenkinsci.maven.plugins.hpi;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.DependencyResolutionRequiredException;
 import org.apache.maven.artifact.resolver.filter.ScopeArtifactFilter;
 import org.apache.maven.model.Resource;
 import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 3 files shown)


sephiroth-j:
@timja , @oleg-nenashev: Do you have time to review this PR?

oleg-nenashev (approved):
Looks good to me

timja (approved):
not tested but seems fine

daniel-beck:
Is a bit light on tests, is that something that could still be added? Otherwise I wouldn't be surprised if we manage to break this at the next opportunity.

sephiroth-j:
@daniel-beck: I was not familiar with those integration tests but I was able the add one that covers packaging webresources (#213).

timja:
This prevents live reloading of files in the webapp directory, see https://github.com/jenkinsci/theme-manager-plugin/pull/75#issuecomment-1048172565

src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
```diff
@@ -98,7 +97,12 @@ public void execute() throws MojoExecutionException, MojoFailureException {
             mainSection.addAttributeAndCheck(new Attribute("Libraries", StringUtils.join(paths, ",")));
 
             // compute Resource-Path entry
-            mainSection.addAttributeAndCheck(new Attribute("Resource-Path",warSourceDirectory.getAbsolutePath()));
+            if (webappDirectory != null && webappDirectory.isDirectory()) {
+                mainSection.addAttributeAndCheck(new Attribute("Resource-Path",webappDirectory.getAbsolutePath()));
```
timja, line 101: Caused https://github.com/jenkinsci/maven-hpi-plugin/issues/299

This changed it from the source to target preventing live reload:

Before:

- WarSourceDir: /Users/timja/code/jenkins/solarized-theme-plugin/src/main/webapp

After:
- WebAppDir path: /Users/timja/code/jenkins/solarized-theme-plugin/target/solarized-theme

PR #235, opened 2021-09-16
Fix `mvn compile` in multimodule projects

Vlatombe:
When running `mvn compile` on a multimodule project containing multiple jenkins plugins depending on each other, build is failing with `Unable to retrieve manifest`.

This is because the current validation requires that a manifest is available and expects a packaged file (required `mvn package` at least).
This is skipping the validation in this particular case, issuing a warning.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/it/compile-multimodule-it/pom.xml
+++ b/src/it/compile-multimodule-it/pom.xml
@@ -1,0 +1,44 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>4.25</version>
+  </parent>
+
+  <groupId>org.jenkins-ci.tools.hpi.its</groupId>
+  <artifactId>compile-multimodule-it-parent</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <packaging>pom</packaging>
+  <properties>
+    <java.level>8</java.level>
+  </properties>
+
+  <modules>
+    <module>plugin1</module>
+    <module>plugin2</module>
+  </modules>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateHpiMojo.java
```diff
@@ -41,17 +42,24 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+                if (mainAttributes.containsKey(hName)) {
+                    return new VersionNumber(mainAttributes.getValue(hName));
+                } else {
+                    throw new MojoExecutionException("Could not find Jenkins Version in manifest for " + artifact);
+                }
             }
+        } else {
+            getLog().warn("Skipping jenkins-core validation for " + artifact + " since we rely on sources and don't have a manifest. Use 'package' goal to get validation");
+            // Assume the version is the same
+            return new VersionNumber(findJenkinsVersion());
```
amuniz, line 62: IIRC this could be missing the core version check, as it is not really returning the dependency version but the current project core version. IMO, this should throw an exception pointing the user to run `mvn package` instead of `mvn compile` (or any goal before `package` in the maven lifecycle).

Proper validation runs on `mvn deploy` so releases are safe, but still... getting `mvn test` to not catch a Jenkins core version mismatch does not sound good.

Vlatombe: `mvn test` checks it through `InjectedTest`, see https://github.com/jenkinsci/maven-hpi-plugin/pull/191

src/it/compile-multimodule-it/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <plugin>
+        <groupId>@project.groupId@</groupId>
+        <artifactId>@project.artifactId@</artifactId>
+        <version>@project.version@</version>
+      </plugin>
+    </plugins>
```
jglick, line 41:
```suggestion
```

Not actually used in the aggregator right?

jglick (approved):

src/it/compile-multimodule-it/pom.xml
Vlatombe: Correct.

PR #246, opened 2021-11-24
Remove `bundle-plugins` goal

basil:
As far as I can tell this is completely dead code. I found no references to it anywhere else in the Jenkins ecosystem.

PR #247, opened 2021-11-24
Maven 3 upgrade

basil:
I got a little further than #23, although `hpi:run` still isn't working. I could use some help debugging that. I did not bother trying to get `BundlePluginsMojo` to compile, hence this PR being downstream of #246.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
@@ -11,7 +11,6 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.codehaus.plexus.archiver.jar.Manifest;
 import org.codehaus.plexus.archiver.jar.Manifest.Attribute;
-import org.codehaus.plexus.archiver.jar.Manifest.Section;
 import org.codehaus.plexus.archiver.jar.ManifestException;
 import org.codehaus.plexus.util.IOUtil;
 
@@ -73,7 +72,7 @@ public class HplMojo extends AbstractJenkinsManifestMojo {
         PrintWriter printWriter = null;
         try {
             Manifest mf = new Manifest();
-            Section mainSection = mf.getMainSection();
+            Manifest.ExistingSection mainSection = mf.getMainSection();
             setAttributes(mainSection);
 
             // compute Libraries entry
@@ -137,6 +136,10 @@ public class HplMojo extends AbstractJenkinsManifestMojo {
         for (MavenArtifact artifact : artifacts) {
             if(jenkinsPlugins.contains(artifact.getId()))
                 continue;   // plugin dependencies
+            if (artifact.getDependencyTrail().size() < 2) {
+                throw new IllegalStateException(
+                        "invalid dependency trail: " + artifact.getDependencyTrail());
+            }
             if(artifact.getDependencyTrail().size() >= 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
                 continue;   // no need to have transitive dependencies through plugins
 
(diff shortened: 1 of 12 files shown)


basil:
I have this _mostly_ working, but there is still a problem with `HplMojo#buildLibraries`: it blows up on https://github.com/jenkinsci/maven-hpi-plugin/blob/2f70c5522e3bf1b60c4821e0b5b07da758c2bac3/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java#L142 because `artifact.getDependencyTrail()` returns a valid trail on the main branch but returns an invalid trail with only one element (which is not expected) with the changes from this PR. That is, this code works on the main branch but fails with the changes from this PR:

https://github.com/jenkinsci/maven-hpi-plugin/blob/2f70c5522e3bf1b60c4821e0b5b07da758c2bac3/src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java#L139-L141

@jglick @Vlatombe I'll keep digging but any help would be appreciated.

src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
```diff
@@ -137,6 +136,9 @@ private void buildLibraries(List<String> paths) throws IOException {
         for (MavenArtifact artifact : artifacts) {
             if(jenkinsPlugins.contains(artifact.getId()))
                 continue;   // plugin dependencies
+            if (artifact.getDependencyTrail().size() < 2) {
+                throw new IllegalStateException("invalid dependency trail");
+            }
             if(artifact.getDependencyTrail().size() >= 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
```
Vlatombe, line 142: It's been wrong for 8 years...

```suggestion
            if(artifact.getDependencyTrail().size() > 1 && jenkinsPlugins.contains(artifact.getDependencyTrail().get(1)))
```

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -783,22 +801,39 @@ public void startJetty() throws MojoExecutionException {
...
+
+            Set<Artifact> artifacts = new LinkedHashSet<>();
+            if (result.getDependencyGraph() != null
+                    && !result.getDependencyGraph().getChildren().isEmpty()) {
+                RepositoryUtils.toArtifacts(
+                        artifacts,
+                        result.getDependencyGraph().getChildren(),
+                        Collections.singletonList(project.getArtifact().getId()),
+                        request.getResolutionFilter());
+            }
```
Vlatombe, line 817: Did you find what was wrong with the maven 3 way of doing things? Bug on maven side? It looks weird that the dependency trail wouldn't be populated without this.

basil:
I got this working… turns out I had to use `ProjectDependenciesResolver` rather than `DependencyResolver` (go figure). Anyway, integration tests are now passing, and I tested this with `mvn verify` and `mvn hpi:run` in two plugins, and all seems to work, so I am marking this ready for review.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
basil: >the maven 3 way of doing things

Does not make sense. The comment to `RunMojo#resolveDependencies` says "Performs the equivalent of "@requireDependencyResolution" mojo attribute", and the implementation of `requiresDependencyResolution` in `LifecycleDependencyResolver` uses `ProjectDependenciesResolver`, so my revised code (_not_ the original) _is_ the "Maven 3 way of doing things".

src/main/java/org/jenkinsci/maven/plugins/hpi/HplMojo.java
basil: I am not touching this pre-existing problem. Instead I added an assertion a line above to make sure the state is as we expect. This assertion passes both before and after this PR.

basil:
@Vlatombe I made some fix-ups to the original changes. Can you please take another look? Thanks!

Vlatombe:
I'm not really able to tell since you force-pushed on your previous changes. Will try on some plugins.

Vlatombe (approved):
LGTM

PR #251, opened 2021-11-26
Ensure implementation and specification entries are present in `MANIFEST.MF`

basil:
Fixes an (unreleased) regression introduced in #247.


Vlatombe:
@basil This makes plugin packaging fail badly (cryptic `Caused by: java.io.IOException: invalid header field`) if the maven `project>description` contains a newline character.

basil:
Sorry, I somehow clicked edit rather than reply. Here is my reply:

> @basil This makes plugin packaging fail badly (cryptic Caused by: `java.io.IOException`: invalid header field) if the maven project>description contains a newline character.

Thanks for the notification. Was this an informational notification, or are you implying that I should take some sort of action as a result of this notification?

> By the way, shouldn't `Specification-Title` be `project.getName()` instead?

You tell me. Should it? Shouldn't it? Why? Why not?

PR #258, opened 2021-12-14
Link to `/doc/developer/plugin-development/mark-a-plugin-incompatible/`

jglick:
@daniel-beck pointed me to this replacement for https://wiki.jenkins.io/display/JENKINS/Marking+a+new+plugin+version+as+incompatible+with+older+versions.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
@@ -60,7 +60,7 @@ public abstract class AbstractJenkinsManifestMojo extends AbstractHpiMojo {
     /**
      * Optional - the oldest version of this plugin which the current version is
      * configuration-compatible with.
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -60,6 +60,7 @@
     /**
      * Optional - the oldest version of this plugin which the current version is
      * configuration-compatible with.
+     * @see <a href="Mark a new plugin version as incompatible with older versions">https://www.jenkins.io/doc/developer/plugin-development/mark-a-plugin-incompatible/</a>
```
daniel-beck, line 63:
```suggestion
     * @see <a href="https://www.jenkins.io/doc/developer/plugin-development/mark-a-plugin-incompatible/">Mark a new plugin version as incompatible with older versions</a>
```

daniel-beck (dismissed):

daniel-beck (approved):

basil (approved):

PR #266, opened 2021-12-19
Use `mojo-executor`

basil:
Addresses this long-standing comment from commit commit ee161c7:

> All I want to do here is to invoke the hpl target. there must be a better way to do this!
> jglick: perhaps https://github.com/TimMoore/mojo-executor would work?

Indeed, it would.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+        <groupId>org.twdata.maven</groupId>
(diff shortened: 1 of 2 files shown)


jglick (approved):
Looks right, and evidently https://github.com/jenkinsci/maven-hpi-plugin/blob/5d3d4cde13fd4ddc6e95f69ab99b484648d54fd0/src/it/parent-3x/invoker.properties#L3 passes.

basil:
I also tested this interactively.

PR #284, opened 2022-01-21
Use name for `Specification-Title` instead of description.

Vlatombe:
Follows up https://github.com/jenkinsci/maven-hpi-plugin/pull/251

Project description ends up being displayed in the plugin manager
(rendered in a html block). It can span multiple lines. Such information
can be long and doesn't really fit in the manifest. Use the name
instead.

Also completed coverage for manifest content.

### DIFF
--- a/src/it/JENKINS-45740-metadata/pom.xml
+++ b/src/it/JENKINS-45740-metadata/pom.xml
@@ -11,6 +11,11 @@
     <artifactId>JENKINS-45740-metadata</artifactId>
     <version>1.0-SNAPSHOT</version>
     <packaging>hpi</packaging>
+    <name>JENKINS-45740 name</name>
+    <description>
+      JENKINS-45740 description.
+      On two lines.
+    </description>
     <properties>
         <java.level>8</java.level>
         <hpi-plugin.version>@project.version@</hpi-plugin.version>
--- a/src/it/JENKINS-45740-metadata/verify.groovy
+++ b/src/it/JENKINS-45740-metadata/verify.groovy
@@ -28,6 +28,13 @@ checkArtifact(
       'Plugin-License-Url': 'https://opensource.org/licenses/MIT',
       'Plugin-License-Name-2': 'MY License',
       'Plugin-License-Url-2': 'https://mylicense.txt',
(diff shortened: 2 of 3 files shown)


jglick (commented):
I suppose this is right although I am not sure what the concrete effect is.

* What, if anything, is `Specification-Title` used for in plugin manifests? I am familiar with its usage in JARs generally.
* Plugins should not normally have a `<description>`. The description should rather be in `src/main/resources/index.jelly`.

basil:
I think using the name rather than the description makes sense in the ideal case. Recent versions of `maven-archiver` use the name [when they fill in `Specification-Title`](https://github.com/apache/maven-archiver/blob/a00e8309132acf45c3e5c81849f51e55dc4fa2e8/src/main/java/org/apache/maven/archiver/MavenArchiver.java#L512) if `isAddDefaultSpecificationEntries` is defined, which is _not_ currently the case in this repository.

The reason I opted to add my own logic for this in #251 rather than enable `isAddDefaultSpecificationEntries` is because of compatibility concerns. As you note, older versions of `maven-hpi-plugin` have historically put the description rather than the name in `Specification-Title` (by virtue of their use of older versions of core Maven functionality), and [core uses this to display the plugin description](https://github.com/jenkinsci/jenkins/blob/0cb248b6bc0152faef34ccd1ac35675e6f55062d/core/src/main/resources/hudson/PluginManager/installed.jelly#L109-L130) (at least when there is no description in `src/main/resources/index.jelly`). So switching `Specification-Title` to the name has the potential to cause a regression in the display of descriptions for plugins whose description is currently defined in `pom.xml` rather than `src/main/resources/index.jelly`. Rather than try to identify such plugins and update them one-by-one, I opted to preserve the existing behavior in #251. But if we have made a good faith effort to ensure that nothing will be broken by this change, I would be in favor of using the name, or even enabling `isAddDefaultSpecificationEntries` to have `maven-artifact` fill in these attributes for us without custom code.

Vlatombe:
Thanks for the core reference. If it is used only as last resort and is not the preferred way, I think the hpi plugin should just discourage users from using the project description altogether.

PR #298, opened 2022-02-22
Fine-grained SpotBugs exclusions

basil:
### Problem

Currently, we exclude any medium-threshold SpotBugs violations. This means that existing violations don't trip up the build, but it also means we won't notice if new medium-threshold violations get introduced.

### Solution

Add an exclude filter file with fine-grained suppressions of specific medium-threshold violations for the classes that currently violate them. This way, the build remains green, but if a new medium-threshold violation gets introduced, we would notice that right away.


basil:
Does anyone object to the merge of this PR? I will wait a day or so before concluding that nobody objects.

Issue #299, opened 2022-02-22
Live reloading of webapp resources broken


timja:
Caused by https://github.com/jenkinsci/maven-hpi-plugin/pull/208

Picked up when updating theme plugins in https://github.com/jenkinsci/theme-manager-plugin/pull/75

To reproduce:

Run this plugin:
https://github.com/jenkinsci/solarized-theme-plugin

Bump parent pom to 4.25

Run with `mvn hpi:run`
Configure the theme on system config page

Change a color, e.g. one in https://github.com/jenkinsci/solarized-theme-plugin/blob/master/src/main/webapp/solarized-dark.css#L3

Refresh the page.

It will get updated.

Bump parent pom to 4.26 (includes https://github.com/jenkinsci/maven-hpi-plugin/pull/208).

Repeat above, file won't be live reloaded

sephiroth-j:
If you only run `mvn clean hpi:run`, the `resource-path` will still point to "src\main\webapp". If the output directory exists, e.g. after packaging, then the output directory will be used for `resource-path`. This is important to support `webResources` and its `targetPath`. In this case, the same structure of the final hpi package will be used to run the plugin locally. You can then edit the files in the output directory instead of the source directory and you will see the changes immediately (_live reloading_).

I am confident that there is another option instead of reverting #208. However, the decision is up to others.

It would certainly be possible to include the various directories of each `webResource` in the list of directories that are monitored for changes. The only thing I do not know yet is how to trigger the output directory update (packaging goal).

jglick:
> I am confident that there is another option instead of reverting #208.

Perhaps, but unless the fix is immediately apparent, we prefer to revert first and leave the amended feature implementation for some unspecified future time. Was there a particular plugin relying on this feature? First I had heard of it.

sephiroth-j:
> Was there a particular plugin relying on this feature? First I had heard of it.

[Dependency-Track Plugin](https://github.com/jenkinsci/dependency-track-plugin/blob/28df2903df4797a220e54057a6596bdef62306bf/pom.xml#L217)

timja:
better to just use something like esbuild or webpack imo.

jglick:
Yeah IIUC there are more standard ways of packaging JS libs in Jenkins plugins.

PR #300, opened 2022-02-23
Revert support targetPath property for webResources

timja:
- Revert "add test for packaging webresources (#213)"
- Revert "support targetPath property for webResources (#208)"

Fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/299

The above change switched the resource path to:

```
target/solarized-theme
```

instead of
```
src/main/webapp
```

Which prevented live reload of webapp resources which is critical when doing UI development.
(Restarting the `hpi:run` isn't enough either you had to run a packaging goal too to work around it)

It _may_ be possible to add multiple directories to the resource path but not sure if having both `src` and `target` will work.

I tried adding an integration test for this but what I had passed both before and after this change 😢 
https://github.com/jenkinsci/maven-hpi-plugin/compare/master...timja:fix-live-reload?expand=1

I've manually tested this and it fixes the issue, full steps are in the issue description (https://github.com/jenkinsci/maven-hpi-plugin/issues/299)


daniel-beck:
Could `stapler.resourcePath` be updated to match?

jglick (approved):
I can cut a release (not sure offhand who else can), so let me know.

timja:
> I can cut a release (not sure offhand who else can), so let me know.

Core team: https://github.com/jenkins-infra/repository-permissions-updater/blob/06c023e956cdc5088fca38d3cb815273d86c3083/permissions/component-maven-hpi-plugin.yml#L8

timja:
Released https://github.com/jenkinsci/maven-hpi-plugin/releases/tag/maven-hpi-plugin-3.25

jglick:
⇒ https://github.com/jenkinsci/plugin-pom/pull/509

PR #302, opened 2022-03-04
Enforce existence of `src/main/resources/index.jelly`

jglick:
See https://github.com/jenkinsci/junit-attachments-plugin/pull/28 for example.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
@@ -24,6 +24,7 @@ import org.apache.maven.archiver.MavenArchiver;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.DependencyResolutionRequiredException;
 import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
 import org.apache.maven.plugins.annotations.Component;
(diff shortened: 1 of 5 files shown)


basil (approved):

basil:
Only 4 of the 23 plugins I maintain needed changes. Not bad!

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -133,6 +134,14 @@ private void performPackaging()
             archiver.setArchiver(jarArchiver);
             archiver.setOutputFile(jarFile);
             jarArchiver.addConfiguredManifest(manifest);
+            File indexJelly = new File(getClassesDirectory(), "index.jelly");
+            if (!indexJelly.isFile()) {
+                throw new MojoFailureException("Missing " + indexJelly + ". Delete any <description> from pom.xml and create src/main/resources/index.jelly:\n" +
```
uhafner, line 139: Why do we want to delete the description from the pom? It makes sense to have a description in the pom as well.

jglick: You do not _have_ to delete `<description>`; it would just be redundant and so there is a chance someone would try to make some text edits expecting them to appear in the product and be confused when they do not.

zbynek:
For the record there are 10 plugins affected https://issues.jenkins.io/browse/JENKINS-68300

PR #315, opened 2022-03-25
Modernize tests

basil:
The tests for this component were using ancient versions of Jenkins core and the plugin parent POM. This PR updates the tests to use relatively recent versions of Jenkins core and the plugin parent POM for more realistic testing.

### DIFF
--- a/src/it/JENKINS-45740-metadata/pom.xml
+++ b/src/it/JENKINS-45740-metadata/pom.xml
@@ -5,13 +5,15 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.47</version>
+        <version>4.38</version>
+        <relativePath />
     </parent>
     <groupId>org.jenkins-ci.tools.hpi.its</groupId>
     <artifactId>JENKINS-45740-metadata</artifactId>
     <version>1.0-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <properties>
+        <jenkins.version>2.249.1</jenkins.version>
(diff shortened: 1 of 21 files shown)


src/it/JENKINS-45740-metadata/pom.xml
```diff
@@ -46,9 +48,9 @@
     </pluginRepositories>
 
     <scm>
-      <connection>scm:git:git://github.com/jenkinsci/${project.artifctId}-plugin.git</connection>
-      <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
-      <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-      <tag>HEAD</tag>
+        <connection>scm:git:git://github.com/jenkinsci/${project.artifctId}-plugin.git</connection>
```
MarkEWaite, line 51: Since GitHub is deprecating git:// protocol, we should use https:// for read access to repositories.
```suggestion
        <connection>scm:git:https://github.com/jenkinsci/${project.artifctId}-plugin.git</connection>
```

MarkEWaite (commented):
I'm not especially credible as a reviewer of Maven plugin changes, but I think this looks good to me.

MarkEWaite (approved):

src/it/JENKINS-45740-metadata/pom.xml
basil: Thanks Mark. Good to update these one by one as we see them, but the whole ecosystem could use a broad sweep. How I wish I had a tool to do that. :smile:

PR #318, opened 2022-03-27
Add default manifest entries with Maven

basil:
Cleans up a small amount of technical debt introduced in #251 during the transition to Maven 3. Supersedes #284. With this PR, we rely on Maven to add the default specification entries and default implementation entries to the manifest. This allows us to remove some ugly copypasta.

This change is not 100% backward-compatible, but it should be safe for the reasons given in the comments.

### DIFF
--- a/src/it/verify-it/pom.xml
+++ b/src/it/verify-it/pom.xml
@@ -17,6 +17,30 @@
     <packaging>hpi</packaging>
 
     <name>MyNewPlugin</name>
+    <description>My New Plugin</description>
+    <url>https://github.com/jenkinsci/maven-hpi-plugin</url>
+
+    <licenses>
+        <license>
+            <name>MIT License</name>
+            <url>https://opensource.org/licenses/MIT</url>
+        </license>
+    </licenses>
+
+    <developers>
+        <developer>
+            <id>nchomsky</id>
+            <name>Noam Chomsky</name>
+            <email>nchomsky@example.com</email>
+        </developer>
+    </developers>
+
+    <scm>
+        <connection>scm:git:https://github.com/jenkinsci/maven-hpi-plugin.git</connection>
+        <developerConnection>scm:git:git@github.com:jenkinsci/maven-hpi-plugin.git</developerConnection>
+        <tag>HEAD</tag>
+        <url>https://github.com/jenkinsci/maven-hpi-plugin</url>
+    </scm>
 
     <properties>
         <jenkins.version>2.249.1</jenkins.version>
--- a/src/it/verify-it/verify.groovy
+++ b/src/it/verify-it/verify.groovy
@@ -17,6 +17,10 @@
  * under the License.
  */
 
+import java.io.InputStream
+import java.nio.file.Files
+import java.util.jar.Manifest
+
 assert new File(basedir, 'target/classes').exists();
 assert new File(basedir, 'target/classes/org/jenkinsci/tools/hpi/its').exists();
 assert new File(basedir, 'target/classes/org/jenkinsci/tools/hpi/its/HelloWorldBuilder.class').exists();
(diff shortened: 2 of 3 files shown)


src/it/verify-it/verify.groovy
```diff
@@ -31,6 +35,34 @@ assert new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools
...
 assert content.contains(" holder.format(\"it.msg\");");
 
+assert new File(basedir, 'target/verify-it/META-INF/MANIFEST.MF').exists()
+
+InputStream is = Files.newInputStream(new File(basedir, 'target/verify-it/META-INF/MANIFEST.MF').toPath())
+try {
+  Manifest manifest = new Manifest(is)
+  assert !manifest.getMainAttributes().getValue('Build-Jdk-Spec').isEmpty()
+  assert manifest.getMainAttributes().getValue('Created-By').startsWith('Maven Archiver')
+  assert manifest.getMainAttributes().getValue('Extension-Name') == null // was provided by Maven 2, but core prefers Short-Name
```
basil, line 45: Maven 2 set this to the short name, but Maven 3 does not. Core always checks `Short-Name` first and only falls back to `Extension-Name` if `Short-Name` is not present. And we always set `Short-Name` in this plugin. So this was always redundant and can be safely removed.

src/it/verify-it/verify.groovy
```diff
@@ -31,6 +35,34 @@ assert new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools
...
+  assert manifest.getMainAttributes().getValue('Group-Id').equals('org.jenkins-ci.tools.hpi.its')
+  assert manifest.getMainAttributes().getValue('Hudson-Version').equals('2.249.1')
+  assert manifest.getMainAttributes().getValue('Implementation-Title').equals('MyNewPlugin') // was project.artifactId in previous versions, now project.name
```
basil, line 48: Maven 2 set this to the artifact ID, but Maven 3 sets this to the project name. I couldn't find any code in the ecosystem that relied on this being set to any particular value.

src/it/verify-it/verify.groovy
```diff
@@ -31,6 +35,34 @@ assert new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools
...
+  assert manifest.getMainAttributes().getValue('Long-Name').equals('MyNewPlugin')
+  assert manifest.getMainAttributes().getValue('Manifest-Version').equals('1.0')
+  assert manifest.getMainAttributes().getValue('Minimum-Java-Version').equals('1.8')
+  assert manifest.getMainAttributes().getValue('Plugin-Developers').equals('Noam Chomsky:nchomsky:nchomsky@example.com')
+  assert manifest.getMainAttributes().getValue('Plugin-License-Name').equals('MIT License')
+  assert manifest.getMainAttributes().getValue('Plugin-License-Url').equals('https://opensource.org/licenses/MIT')
+  assert manifest.getMainAttributes().getValue('Plugin-ScmUrl').equals('https://github.com/jenkinsci/maven-hpi-plugin')
+  assert manifest.getMainAttributes().getValue('Plugin-Version').startsWith('1.0-SNAPSHOT')
+  assert manifest.getMainAttributes().getValue('Short-Name').equals('verify-it')
+  assert manifest.getMainAttributes().getValue('Specification-Title').equals('MyNewPlugin') // was project.description in previous versions, now project.name
```
basil, line 60: Core does consume this to display on the installed plugins page, but only if `src/main/resources/index.jelly` is not present. As of #302 we enforce the existence of `src/main/resources/index.jelly`, so changing the semantics of `Specification-Title` will not result in any change in behavior.

Vlatombe (approved):

src/it/verify-it/pom.xml
```diff
@@ -17,6 +17,30 @@
...
+            <name>Noam Chomsky</name>
+            <email>nchomsky@example.com</email>
+        </developer>
+    </developers>
+
+    <scm>
+        <connection>scm:git:https://github.com/jenkinsci/maven-hpi-plugin.git</connection>
+        <developerConnection>scm:git:git@github.com:jenkinsci/maven-hpi-plugin.git</developerConnection>
+        <tag>HEAD</tag>
+        <url>https://github.com/jenkinsci/maven-hpi-plugin</url>
```
jglick, line 42: ditto here

jglick (approved):

src/it/verify-it/pom.xml
basil: Done

src/it/verify-it/verify.groovy
```diff
@@ -31,6 +35,34 @@ assert new File(basedir, 'target/generated-sources/localizer/org/jenkinsci/tools
...
+  assert manifest.getMainAttributes().getValue('Url').equals('https://github.com/jenkinsci/maven-hpi-plugin')
+} finally {
+  is.close()
+}
```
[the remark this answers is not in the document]
basil: Done

PR #319, opened 2022-03-29
Improve validation mojo

basil:
The validation mojo currently checks for Java 6 or later and Jenkins 1.420 or later. These seem like pretty old minimum requirements.

This PR bumps the minimum supported core version to 2.204 and enforces minimum Java version based on the value encoded by core, at whatever version of core the plugin is running against. The only place this is encoded right now is in `java.level` in `core-bom`, which is where we're reading it (well, `MANIFEST.MF` in `jenkins.war` encodes a `Build-Jdk-Spec`, but that could be problematic if you did a local build with Java 11 and `<release>8</release>`, so I am not relying on it). I don't feel strongly about encoding it in this particular way, just that it is encoded _somewhere_ in the artifacts published by core and that we read it out here. If we want to encode it some other way I am fine with that.

I tested this by building core with a `java.level` of 11 and verifying that a plugin gave us the proper error message. No integration tests yet, because it is too hard to integration test all of this stuff across multiple repositories when stuff hasn't been released. When releases have been made and everything is updated, I think it will be far easier to add test coverage.

@jglick Curious what you think in the context of our other discussions elsewhere. No pressure to review this though.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
     <dependency>
       <groupId>org.jenkins-ci</groupId>
       <artifactId>version-number</artifactId>
-      <version>1.10-rc142.172b_43188179</version> <!-- TODO https://github.com/jenkinsci/lib-version-number/pull/41 -->
+      <version>1.10</version>
     </dependency>
     <dependency>
       <groupId>org.kohsuke.stapler</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,7 @@
     <dependency>
       <groupId>org.jenkins-ci</groupId>
       <artifactId>version-number</artifactId>
-      <version>1.9</version>
+      <version>1.10-rc142.172b_43188179</version> <!-- TODO https://github.com/jenkinsci/lib-version-number/pull/41 -->
     </dependency>
     <dependency>
       <groupId>org.kohsuke.stapler</groupId>
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
@@ -2,7 +2,12 @@ package org.jenkinsci.maven.plugins.hpi;
 
 import hudson.util.VersionNumber;
 import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
-import java.util.Properties;
+import java.io.DataInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.factory.ArtifactFactory;
 import org.apache.maven.execution.MavenSession;
@@ -14,7 +19,6 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.MavenProjectHelper;
 import org.apache.maven.project.ProjectBuilder;
-import org.apache.maven.project.ProjectBuildingException;
 import org.apache.maven.shared.transfer.artifact.DefaultArtifactCoordinate;
 import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolver;
 import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolverException;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
```diff
@@ -89,6 +94,39 @@ protected String findJenkinsVersion() throws MojoExecutionException {
...
+        try {
+            properties = wrap(bom).resolvePom().getProperties();
+        } catch (ProjectBuildingException e) {
+            throw new MojoExecutionException("Failed to resolve artifact " + bom, e);
+        }
+
+        String version = properties.getProperty("java.level");
+        if (version == null || version.isEmpty()) {
+            throw new MojoExecutionException("java.level not defined in " + bom);
+        }
```
basil, line 110: This is probably the most controversial part of this change, as it commits us to an encoding of `java.level` in the `<properties>` section of `core-bom`. Might not be the ideal encoding. As I wrote above, my point here is just to rely on an encoding _somewhere_. It doesn't have to be this specific encoding.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
```diff
@@ -89,6 +94,39 @@ protected String findJenkinsVersion() throws MojoExecutionException {
...
+        artifactCoordinate.setArtifactId("jenkins-bom");
+        artifactCoordinate.setVersion(version);
+        artifactCoordinate.setExtension("pom");
+
+        try {
+            return artifactResolver
+                    .resolveArtifact(session.getProjectBuildingRequest(), artifactCoordinate)
+                    .getArtifact();
+        } catch (ArtifactResolverException e) {
+            throw new MojoExecutionException("Couldn't download artifact: ", e);
```
jglick, line 126: IIUC this will be the failure point if your `jenkins.version` predates the version where the core BOM was introduced? So it would be polite to make the error message just say what that minimum version is.

basil: Ah I see your point. Well the next check takes care of that by requiring 2.204 or later. So I can just reverse the order of the checks to make the message more polite.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
jglick: Can we inspect the Java bytecode level of some class in `jenkins-core.jar`? More work, but less sensitive to how core is built.

jglick: e.g. https://stackoverflow.com/a/1707200/12916 / https://stackoverflow.com/a/1293372/12916

basil: That's an interesting idea. I'll try it out. I like it on a high level. As the late Roger Faulkner once opined:

> Roger believed that terrible things were sometimes required to create beautiful abstractions, and his trailblazing work on `/proc` embodies this burden: the innards may be delicate and nasty (‘vile,’ as Roger might say in his distinguished Carolinian accent) — but the resulting abstractions are breathtaking in their power, scope and robustness.

basil: Done in commit 34cf73b9736fb1b619bc148f5a9fca45528bb610. Marking as draft due to the incremental in the POM file, but from my perspective this is ready for review, just not ready for merge until the release of `lib-version-number`.

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
```diff
@@ -13,24 +13,16 @@
...
-    /**
-     * Are we running on JDK6 or above?
-     */
-    private static boolean isMustangOrAbove() {
-        try {
-            Class.forName("javax.annotation.processing.Processor");
-            return true;
-        } catch(ClassNotFoundException e) {
-            return false;
+        if (new VersionNumber(findJenkinsVersion()).compareTo(new VersionNumber("2.203.99")) <= 0) {
```
jglick, line 24: Why not
```suggestion
        if (new VersionNumber(findJenkinsVersion()).compareTo(new VersionNumber("2.204")) < 0) {
```
?

jglick (approved):
Looks good. Do you also plan to verify in a mojo that the plugin’s Java release is the same as that of core?

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
basil: The same question could have been asked of the original code, but no particular reason. Trivial enough to fix while I am here I suppose.

pom.xml
```diff
@@ -163,7 +163,7 @@
     <dependency>
       <groupId>org.jenkins-ci</groupId>
       <artifactId>version-number</artifactId>
-      <version>1.9</version>
+      <version>1.10-rc142.172b_43188179</version> <!-- TODO https://github.com/jenkinsci/lib-version-number/pull/41 -->
```
jglick, line 166:
```suggestion
      <version>1.10</version>
```

basil:
> Do you also plan to verify in a mojo that the plugin’s Java release is the same as that of core?

I need to think about this some more.

basil:
> Do you also plan to verify in a mojo that the plugin’s Java release is the same as that of core?

In https://github.com/jenkinsci/maven-hpi-plugin/pull/323 I am not verifying it, but rather dynamically adjusting it as needed to allow for a smoother transition. Once the majority of plugins have crossed the flag day, we can likely change this from a dynamic adjustment to a hard enforcement, but this makes the transition smoother.

PR #320, opened 2022-03-30
Print warning if using deprecated `java.level` with new enough parent POM

basil:
Tested manually with https://github.com/jenkinsci/plugin-pom/pull/522


timja (approved):

basil:
I couldn't find a way to do this with a POM profile. Enforcer has a [`requireProperty`](https://maven.apache.org/enforcer/enforcer-rules/requireProperty.html) rule, but here we want to do the opposite (ban a property).

jetersen (approved):

basil:
> I think you could do this with a POM profile

Actually I think not. From [the Maven documentation on profiles](https://maven.apache.org/guides/introduction/introduction-to-profiles.html):

> Supported variables are system properties like `${user.home}` and environment variables like `${env.HOME}`. Please note that properties and values defined in the POM itself are not available for interpolation here.

So yeah this is just not possible.

PR #322, opened 2022-03-30
Dynamically apply `Add-Opens` directives from `MANIFEST.MF`

basil:
Currently, people testing with Java 17 must add an awkward `-DargLine=` directive with the `Add-Opens` modules from the core version they are testing. This PR applies a trick to dynamically set this based on the values in `MANIFEST.MF`.

With this in place, anyone with a sufficiently recent core (2.341+), test harness (1721.v385389722736+), and `maven-hpi-plugin` (this PR) should be able to effortlessly do Java 17 testing and indeed enable Java 17 in their `Jenkinsfile`. Once this is merged, I will try to do a release and wrap this and the test harness change into a new plugin POM release so that the requirement just becomes "core 2.341+ and plugin POM 4.40+".

Tested with `workflow-job` and Java 17:

```
$ mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.job.WorkflowJobTest,org.jenkinsci.plugins.workflow.job.WorkflowRunTest -Djenkins.version=2.341 -Djenkins-test-harness.version=1723.vcd938b_e66072 -Dhpi-plugin.version=3.27-SNAPSHOT -Denforcer.skip
[…]
[INFO] --- maven-hpi-plugin:3.27-SNAPSHOT:initialize (default-initialize) @ workflow-job ---
[INFO] Setting argLine to -Xms768M -Xmx768M -XX:+HeapDumpOnOutOfMemoryError -XX:+TieredCompilation -XX:TieredStopAtLevel=1 --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED
[…]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:14 min
[INFO] Finished at: 2022-03-30T11:21:40-07:00
[INFO] ------------------------------------------------------------------------
$
```

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
@@ -1,5 +1,6 @@
 package org.jenkinsci.maven.plugins.hpi;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import hudson.util.VersionNumber;
 import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
 import java.io.DataInputStream;
@@ -7,6 +8,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.jar.JarFile;
+import java.util.jar.Manifest;
 import java.util.zip.ZipEntry;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.factory.ArtifactFactory;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
```diff
@@ -0,0 +1,55 @@
...
+            return;
+        }
+
+        String addOpens = getAddOpens();
+        if (addOpens == null) {
+            // core older than 2.339, ignore
+            return;
+        }
+
+        String argLine = project.getProperties().getProperty("argLine");
```
jglick, line 35: This is not great—it ties us to using a property of this name in `plugin-pom`. (I am also not sure whether this works correctly when `argLine` is overridden with references to other properties. Worth checking `-Penable-jacoco` specifically.)

Better would be for this mojo to define a fresh property, `jenkins.addOpens` or whatever, and adjust `plugin-pom` to use it:

```xml
<jenkins.addOpens/>
<argLine>${jenkins.addOpens} -Xms768M -etc</argLine>
```

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsMojo.java
```diff
@@ -140,6 +142,37 @@ private Artifact resolveJenkinsCore() throws MojoExecutionException {
         }
     }
 
+    @CheckForNull
+    protected String getAddOpens() throws MojoExecutionException {
```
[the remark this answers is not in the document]
basil: Fine, commit 4150f31

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
basil: > This is not great—it ties us to using a property of this name in `plugin-pom`.

And how is this not great?

> (I am also not sure whether this works correctly when argLine is overridden with references to other properties. Worth checking -Penable-jacoco specifically.)

Have you proven this, or this idle speculation?

> Better would be for this mojo to define a fresh property, `jenkins.addOpens` or whatever, and adjust `plugin-pom` to use it

commit 0e11da2

https://github.com/jenkinsci/plugin-pom/pull/524

jglick (approved):
Nice. BTW how does this sort of thing get integration-tested? Does running existing `plugin-pom` ITs using Java 17 suffice?

basil:
> Nice. BTW how does this sort of thing get integration-tested? Does running existing plugin-pom ITs using Java 17 suffice?

I will write some tests once all these changes are merged and dependencies are updated. Right now there are too many dependencies on unmerged/unreleased PRs to be able to write tests easily.

PR #323, opened 2022-03-30
Dynamically configure Java compiler based on core

basil:
When combined with https://github.com/jenkinsci/plugin-pom/pull/523 (which this PR doesn't _require_, but rather _benefits from_ if present), this dynamically configures the Java compiler settings based on the core version in use. Although this sort of dynamic configuration may confuse IDEs, the benefit is that it avoids an awkward flag day when we bump the minimum required Java version to 11. Of course, we may still choose to update the plugin parent POM at that point in time for the benefit of IDEs, but having this change widely adopted beforehand would soften the blow.

I've tested this with a minimum Java version of 8 and a minimum Java version of 11, with a version 8 compiler and a version 11 compiler (matrix of 4 test configurations). When using a minimum Java version of 11, I verified that I could use Java 11 features in the plugin. When using a minimum Java version of 11 and a Java 8 compiler, I verified that I got an error. When using a Java 11 compiler, I verified that we were only setting `release` and `testRelease` regardless of whether the minimum Java version was 8 or 11. When using a Java 8 compiler, I verified that we were setting `source`, `target`, `testSource`, and `testTarget` but not `release` or `testRelease`.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
@@ -1,0 +1,75 @@
+package org.jenkinsci.maven.plugins.hpi;
+
+import hudson.util.VersionNumber;
+import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+
+/**
+ * Configure Maven for the desired version of Java.
+ *
+ * @author Basil Crow
+ */
+@Mojo(name = "initialize", defaultPhase = LifecyclePhase.INITIALIZE)
+public class InitializeMojo extends AbstractJenkinsMojo {
+
+    @Override
+    public void execute() throws MojoExecutionException {
+        setCompilerProperties();
+    }
+
+    private void setCompilerProperties() throws MojoExecutionException {
+        if (!project.getProperties().containsKey("maven.compiler.source")
+                && !project.getProperties().containsKey("maven.compiler.release")) {
+            // On an older plugin parent POM that predates the setting of these values as Maven properties.
+            return;
+        }
+
+        JavaSpecificationVersion javaVersion = getMinimumJavaVersion();
+        if (JavaSpecificationVersion.forCurrentJVM().isOlderThan(new VersionNumber("9"))) {
+            // Should always be set already, but just in case...
+            setProperty("maven.compiler.source", javaVersion.toString());
+            setProperty("maven.compiler.target", javaVersion.toString());
+            setProperty("maven.compiler.testSource", javaVersion.toString());
+            setProperty("maven.compiler.testTarget", javaVersion.toString());
+            // Should never be set already, but just in case...
+            unsetProperty("maven.compiler.release");
+            unsetProperty("maven.compiler.testRelease");
+        } else {
+            /*
+             * When compiling with a Java 9+ compiler, we always rely on "release" in favor of "source" and "target",
+             * even when compiling to Java 8 bytecode.
+             */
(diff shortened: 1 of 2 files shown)


basil:
Sample output from "migration mode" (minimum Java version faked to 11), compiling with Java 11:

```
maven-hpi-plugin:3.27-SNAPSHOT:initialize (default-initialize) @ pipeline-stage-tags-metadata ---
[INFO] Setting maven.compiler.release to 11
[INFO] Setting maven.compiler.testRelease to 11
[INFO] Unsetting maven.compiler.source
[INFO] Unsetting maven.compiler.target
[INFO] Unsetting maven.compiler.testSource
[INFO] Unsetting maven.compiler.testTarget
```

The intention is that when we do decide to drop support for Java 8, we would merge and release https://github.com/jenkinsci/plugin-pom/pull/478 which would eliminate the above sketchy dynamic adjustments, but in the meantime this helps make the transition smoother.

timja (commented):
This worries me a bit as too magical.

What's wrong with just requiring a plugin pom update to build and compile Java 11+?

Plugins should still compile fine on Java 8 (as long as they're built against older cores), and will be able to run fine on Java 11 and newer cores.

----

Is this to allow compiling with Java release level 11 in CI before upgrading the pom to a version that requires Java 11?

----

(Non blocking to be clear)

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
```diff
@@ -0,0 +1,75 @@
...
+public class InitializeMojo extends AbstractJenkinsMojo {
+
+    @Override
+    public void execute() throws MojoExecutionException {
+        setCompilerProperties();
+    }
+
+    private void setCompilerProperties() throws MojoExecutionException {
+        if (!project.getProperties().containsKey("maven.compiler.source")
+                && !project.getProperties().containsKey("maven.compiler.release")) {
```
jglick, line 24: This presumes that the POM is configuring these mojo parameters via properties, which is not necessary, and in fact discouraged—best configured via `<configuration>` in `<plugin>`.

More robust to look up the effective configuration of the mojo. Do not recall the idiom for that offhand.

jglick:
> What's wrong with just requiring a plugin pom update to build and compile Java 11+?

And not _allowing_ you to the newer POM to build and compile at level 8—so that would be a breaking update for the plugin POM. I also think this would be acceptable given the special circumstances. https://github.com/jenkinsci/plugin-pom/pull/478#issuecomment-1082198091 has more detailed discussion. I guess I am -0 on this—seems like asking for trouble, and not exactly necessary, but also not blocking either (since this magic would not be expected to be used most of the time).

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
basil: > More robust to look up the effective configuration of the mojo. Do not recall the idiom for that offhand.

Declined. If you want to understand the reason why then just try writing the code and you will quickly find out.

basil:
> I guess I am -0 on this

Comments like this really confuse me. In the other PR, you wrote about how breaking compatibility was a concern and how we should try to find ways to avoid a flag day. Well, this is it: signed, sealed, and delivered. And it took a considerable amount of effort to reach this solution. And now you are saying "I'd rather we not do this." No matter which way I attempt to address your concerns, it seems that I cannot win.

Yes this is temporary glue code. Most backward compatibility shims are not pretty. This code can be removed later on, when Mark breaks his silence about a date and we cross over into a hypothetical glorious future by dropping support for Java 8 in `plugin-pom`. I think that a smoother transition for developers is worth the added complexity in the short term.

timja (approved):

jglick (approved):
Makes me nervous, but so long as this is considered a temporary trick it is probably worth it to avoid friction during the transitional period.

basil:
Thanks. And I think it should go without saying at this point that if there are any problems that I will take responsibility for them in a timely fashion, up to and including a revert if necessary.

PR #324, opened 2022-03-30
[JENKINS-67974] Java 11 support for MemoryAssert#assertGC

basil:
Downstream of #322. Upstream of jenkinsci/jenkins-test-harness#400. Still in draft state pending the resolution of apache/netbeans#3743. Tested in Java 17 with `mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest -Djenkins.version=2.341 -Dhpi-plugin.version=3.27-SNAPSHOT`.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
@@ -1,16 +1,28 @@
 package org.jenkinsci.maven.plugins.hpi;
 
-import com.google.common.collect.Iterables;
+import com.google.common.io.ByteStreams;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.util.VersionNumber;
 import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
+import java.util.jar.JarFile;
+import java.util.jar.Manifest;
+import java.util.zip.ZipEntry;
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.ResolutionScope;
-import org.apache.tools.ant.util.StringUtils;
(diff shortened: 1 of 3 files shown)


basil:
Ready, tested with Java 17 and my other PRs with `mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest -Djenkins.version=2.341 -Dhpi-plugin.version=3.27-SNAPSHOT`

Needs a change to the test harness to actually start bundling the hook, and upgrade insane, and get the integration tests to pass. I am preparing that PR next.

basil:
> I am preparing that PR next.

Done in https://github.com/jenkinsci/jenkins-test-harness/pull/416

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
```diff
@@ -22,13 +29,14 @@
  *
  * @author Basil Crow
  */
-@Mojo(name = "initialize", defaultPhase = LifecyclePhase.INITIALIZE)
+@Mojo(name = "initialize", requiresDependencyResolution = ResolutionScope.TEST, defaultPhase = LifecyclePhase.INITIALIZE)
```
jglick, line 32: I think it would make more sense to do this in a test preparation phase if we can?

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
```diff
@@ -144,4 +152,46 @@ private static String buildArgLine(String manifestEntry) {
...
+
+    @NonNull
+    private static Path getInsaneHook(MavenArtifact artifact) throws MojoExecutionException {
+        File jar = artifact.getFile();
+        try (JarFile jarFile = new JarFile(jar)) {
+            ZipEntry entry = jarFile.getEntry("netbeans/harness/modules/ext/org-netbeans-insane-hook.jar");
+            if (entry == null) {
+                throw new MojoExecutionException("Failed to find org-netbeans-insane-hook.jar in " + jar);
+            }
+            Path tempFile = Files.createTempFile("org-netbeans-insane-hook", ".jar");
```
jglick, line 188: Can we set this to delete-on-exit?

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
basil: I agree. I might split this out into a different mojo soonish in some separate PR.

src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
basil: Yes, good idea

basil: Done in commit b654ee8

jglick (approved):

PR #345, opened 2022-06-02
Handle presence of index.jelly differently

twasyl:
Signed-off-by: Thierry Wasylczenko <twasyl@users.noreply.github.com>

Currently if there is no `src/main/resources/index.jelly` file in the Jenkins plugin sources, `mvn hpi:run` will fail the build indicating the file must be created using the POM's `<description>`. This behaviour has been introduced in #302. As I like to enforce the creation of `index.jelly`, the current approach:
 - assumes there is a `<description>` in the POM
 - lead to think you have to delete the `<description>` from your POM

IMO this could be improved for plugins author's UX. If the description is present but the the file is not, I assume the author:
- doesn't need to create a file when it's content is predictive (because we have the description)
- doesn't want the content of `index.jelly` to be different from the description

This PR aims to provide the following behaviours:
- In case the `index.jelly` file doesn't exist and the description is provided, create the file under `target/classes` and log a warning
- In case the `index.jelly` file doesn't exist and the description **is not** provided, then throw an error to enforce the creation of the file by the author
- In case the `index.jelly` file is provided, use it

- [x] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/src/it/missing-index/verify.groovy
+++ b/src/it/missing-index/verify.groovy
@@ -17,6 +17,6 @@
  * under the License.
  */
 
-assert new File(basedir, 'build.log').getText('UTF-8').contains("create src/main/resources/index.jelly:");
+assert new File(basedir, 'build.log').getText('UTF-8').contains("Create src/main/resources/index.jelly:");
 
 return true;
--- a/src/it/missing-index/verify.groovy
+++ b/src/it/missing-index/verify.groovy
@@ -18,7 +18,8 @@
  */
 
 assert new File(basedir, 'build.log').getText('UTF-8').contains("A default one will be created using the description of the pom.xml");
-assert new File(basedir, "target/classes/index.jelly").getText('UTF-8') == """<?jelly escape-by-default='true'?>
+assert new File(basedir, "target/classes/index.jelly").getText('UTF-8') == """<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
 <div>
 Deprecated spot for plugin &quot;description&quot; with &lt;4 special chars.
 </div>"""
--- a/src/it/missing-index/verify.groovy
+++ b/src/it/missing-index/verify.groovy
@@ -18,5 +18,9 @@
  */
 
 assert new File(basedir, 'build.log').getText('UTF-8').contains("A default one will be created using the description of the pom.xml");
+assert new File(basedir, "target/classes/index.jelly").getText('UTF-8') == """<?jelly escape-by-default='true'?>
+<div>
+Deprecated spot for plugin &quot;description&quot; with &lt;4 special chars.
+</div>"""
 
 return true;
--- a/src/it/missing-index/verify.groovy
+++ b/src/it/missing-index/verify.groovy
@@ -17,6 +17,6 @@
  * under the License.
  */
 
-assert new File(basedir, 'build.log').getText('UTF-8').contains("src/main/resources/index.jelly does not exist. A default one will be created using the description of the pom.xml");
+assert new File(basedir, 'build.log').getText('UTF-8').contains("A default one will be created using the description of the pom.xml");
 
 return true;
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
@@ -149,7 +149,7 @@ public class HpiMojo extends AbstractJenkinsManifestMojo {
                          final OutputStreamWriter indexJellyWriter = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
                         indexJellyWriter.write("<?jelly escape-by-default='true'?>\n" +
                                 "<div>\n" +
-                                StringEscapeUtils.escapeHtml(projectDescription) + "\n" +
+                                StringEscapeUtils.escapeXml(projectDescription) + "\n" +
                                 "</div>");
                         indexJellyWriter.flush();
                     }
@@ -157,7 +157,7 @@ public class HpiMojo extends AbstractJenkinsManifestMojo {
                     throw new MojoFailureException("Missing " + indexJelly + ". Create src/main/resources/index.jelly:\n" +
                             "<?jelly escape-by-default='true'?>\n" +
                             "<div>\n" +
-                            "    The description here...\n" +
+                            "    The description here…\n" +
                             "</div>");
                 }
             }
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
@@ -33,6 +33,7 @@ import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
+import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.archiver.Archiver;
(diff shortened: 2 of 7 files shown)


Vlatombe:
I'm not at ease thinking that `hpi:hpi` can somehow write under `src/` directory.

Such goal should always be reproducible, and here if you call it twice, the first call will fail, and the second one will succeed. 

Adding a separate goal to implement such behaviour would be acceptable IMO.

daniel-beck:
I'm not sure generating sources outside directories expected to contain something like that is expected. Since users will need to `git add` the file and commit it, there's no actual automation here.

I could see an argument for generating the `index.jelly` from POM description in the `target/` though, OTOH a case could be made that developers are expected to provide a better description, e.g. (limited) HTML formatting is possible here.

twasyl:
Sorry for maybe being unclear: the generated file is in `target/classes`, not in any source directory.

daniel-beck:
> Sorry for maybe being unclear: the generated file is in `target/classes`, not in any source directory, [see](https://github.com/jenkinsci/maven-hpi-plugin/pull/345/files#diff-82989754ad96788b6431a904da3f182ff71cee3ac880e27dac0aa1555d4a5f2bR144)

Ah, I was confused by the path in the error message 🤦

Vlatombe:
> Sorry for maybe being unclear: the generated file is in `target/classes`, not in any source directory, [see](https://github.com/jenkinsci/maven-hpi-plugin/pull/345/files#diff-82989754ad96788b6431a904da3f182ff71cee3ac880e27dac0aa1555d4a5f2bR144)

I got confused as well. Looks okay to me.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -136,11 +140,23 @@ private void performPackaging()
...
-                    "<div>\n" +
-                    "    The description here…\n" +
-                    "</div>");
+                if (getProjectDescription() != null) {
+                    getLog().warn("src/main/resources/index.jelly does not exist. A default one will be created using the description of the pom.xml");
+                    try (final FileOutputStream fos = new FileOutputStream(indexJelly);
+                         final OutputStreamWriter indexJellyWriter = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
+                        indexJellyWriter.write("<?jelly escape-by-default='true'?>\n" +
+                                "<div>\n" +
+                                project.getDescription() + "\n" +
```
jglick, line 149: You would need to escape any XML metacharacters here.

jglick (commented):
No strong opinion, but

> a case could be made that developers are expected to provide a better description

I think this is a good opportunity to just have the developer take a couple minutes to write the description they wish to show to users, possibly with formatting. Most plugins already have an `index.jelly` so this should only affect a small minority anyway.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
twasyl: Some metacharacters like `&`, `<` and `>` can't be put in the POM's `<description>` as the build will failed.

For other ones, like `"` and `'`, the `StringEscapeUtils#escapeXml(String)` doesn't replace them with their equivalent. So not sure if you are thinking about something else.

src/it/missing-index/verify.groovy
```diff
@@ -17,6 +17,6 @@
  * under the License.
  */
 
-assert new File(basedir, 'build.log').getText('UTF-8').contains("create src/main/resources/index.jelly:");
+assert new File(basedir, 'build.log').getText('UTF-8').contains("A default one will be created using the description of the pom.xml");
```
jglick, line 20: More interesting to assert text of the generated file.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
jglick: > can't be put in the POM's `<description>`

Escaped they can, and `Project.getDescription` will return the raw text.

src/it/missing-index/verify.groovy
twasyl: Yes indeed. I updated the test to also include an assertion on the file's content

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
twasyl: Got it. Updated

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -149,15 +149,15 @@ private void performPackaging()
                          final OutputStreamWriter indexJellyWriter = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
                         indexJellyWriter.write("<?jelly escape-by-default='true'?>\n" +
                                 "<div>\n" +
-                                StringEscapeUtils.escapeXml(projectDescription) + "\n" +
+                                StringEscapeUtils.escapeHtml(projectDescription) + "\n" +
```
jglick, line 152: This is XML, not HTML.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -149,15 +149,15 @@ private void performPackaging()
...
                                 "</div>");
                         indexJellyWriter.flush();
                     }
                 } else {
                     throw new MojoFailureException("Missing " + indexJelly + ". Create src/main/resources/index.jelly:\n" +
                             "<?jelly escape-by-default='true'?>\n" +
                             "<div>\n" +
-                            "    The description here…\n" +
+                            "    The description here...\n" +
```
jglick, line 160: Go back to `…` and make sure you are setting the XML encoding to UTF-8.

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
twasyl: Fixed

src/main/java/org/jenkinsci/maven/plugins/hpi/HpiMojo.java
```diff
@@ -149,15 +149,15 @@ private void performPackaging()
                          final OutputStreamWriter indexJellyWriter = new OutputStreamWriter(fos, StandardCharsets.UTF_8)) {
                         indexJellyWriter.write("<?jelly escape-by-default='true'?>\n" +
```
jglick, line 150: This should be setting an XML encoding line.

```xml
<?xml version="1.0" encoding="UTF-8"?>
```

twasyl: The previous error suggested:

```
throw new MojoFailureException("Missing " + indexJelly + ". Delete any <description> from pom.xml and create src/main/resources/index.jelly:\n" +
                    "<?jelly escape-by-default='true'?>\n" +
                    "<div>\n" +
                    "    The description here…\n" +
                    "</div>");
```
🤣

twasyl: Fixed it

jtnord (commented):
I am not sure why we would want this behaviour at all.

the Update center will use the description if the index.jelly is not found (as will the plugin manager if I am not mistaken).

So all this does is make a currently failing (has a description but no index.jelly) build pass.

You could just as easily do that by not failing in the first place (change the failure to a warning if there is no index.jelly) and avoid this generation.

but if the point is that people do not pay attention to warnings - why would they pay attention to the new warning here and provide something better?

twasyl:
Closing as stale. Nobody approved it in a year.

PR #368, opened 2022-07-27
Relax release check

basil:
This gets in the way of PCT's `install` goal specified in this code path:

https://github.com/jenkinsci/plugin-compat-tester/blob/f035412416dd448b85b472c65202ac6a6b7c50ef/plugins-compat-tester/src/main/java/org/jenkins/tools/test/hook/MultiParentCompileHook.java#L124-L130


jetersen (approved):
🐛

PR #375, opened 2022-08-09
Simplify semantics

basil:
If I am understanding #373 then there is no real use case where one would ever want to set `overrideWar` but not also `jth.jenkins-war.path`, so let's just consume the `jth.jenkins-war.path` property directly the same way the test harness does instead of forcing the user to specify two options both pointing to the same file. With that semantic in place, there is no need to add dependencies not already present in the project model to the project model, since the use case of testing plugin-to-plugin interaction of "long tail" (unrelated) plugins is handled by setting `jth.jenkins-war.path`, and such a use case probably doesn't exist for someone who is setting `overrideVersions` but not `jth.jenkins-war.path`.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
@@ -492,7 +492,6 @@ public class TestDependencyMojo extends AbstractHpiMojo {
             Log log)
             throws MojoExecutionException {
         Set<String> appliedOverrides = new HashSet<>();
-        Set<String> appliedBundledPlugins = new HashSet<>();
 
         // Update existing dependency entries in the model.
         for (Dependency dependency : project.getDependencies()) {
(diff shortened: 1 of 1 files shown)


basil (commented):
The only downside to these semantics would be if there are any consumers that would not want to get this behavior when passing `jth.jenkins-war.path` (unlikely I think) or if they would not want to pass `jenkins.version` which is required when overriding from a WAR in `maven-hpi-plugin` (maybe slightly more likely but probably OK?)

jglick:
> there is no real use case where one would ever want to set `overrideWar` but not also `jth.jenkins-war.path`

There is a use case, and CloudBees has it: we wish to pick up plugin versions from a WAR (hence `overrideWar`) but setting `jth.jenkins-war.path` causes all sorts of complicated failures.

jglick (changes requested):
This change would prevent CloudBees from adopting the new PCT mode.

basil:
Would have been nice to simplify the API, but cannot do so while still supporting all existing consumers.

PR #376, opened 2022-08-09
Slightly simplify semantics

basil:
A middle ground between what I feel are the overly conservative semantics from #373 and the perhaps too risky semantics of #375. No risk to anyone who is passing `jth.jenkins-war.path` but doesn't want any change in behavior or to be forced to pass `jenkins.version`. For those who are passing `jth.jenkins-war.path` and want an override, they now would pass `overrideWar=true`, which cues us to pick up the WAR from `jth.jenkins-war.path`. In that case, the use of `jth.jenkins-war.path` satisfies the use case of testing interaction between the "long tail" of plugins unrelated to the main plugin being tested. That use case doesn't exist for someone who is just using `overrideVersions`, but that consumer probably isn't passing in `overrideWar` anyway.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
@@ -98,14 +98,15 @@ public class TestDependencyMojo extends AbstractHpiMojo {
     private List<String> overrideVersions;
 
     /**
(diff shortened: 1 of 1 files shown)


jglick (changes requested):
This change would prevent CloudBees from adopting the new PCT mode.

basil:
Would have been nice to simplify the API, but cannot do so while still supporting all existing consumers.

PR #377, opened 2022-08-10
Stop adding unrelated dependencies to the model

basil:
Simpler version of #373 that just removes this functionality rather than disabling it by default. Only impact to consumers is for `jenkinsci/bom`, which I will test with the incremental build from this PR.


basil:
I still want to get this done eventually, but I need to do some further work in `jenkinsci/bom` to prepare for this.

basil:
Closing in favor of #461.

PR #379, opened 2022-08-12
Add a `dependencyManagement` section if missing

jglick:
Appears to be necessary to run PCT on `warnings-ng` at least in the CloudBees configuration. (Not clear if https://github.com/jenkinsci/bom/pull/1384 even got that far.)

Amends #336.


jtnord (approved):
code change in isolation seems completely sane to me.

PR #391, opened 2022-11-07
resolve-test-dependencies fail to update classpath to use some optional transitive dependencies coming from hpi included in overrideWar

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>


olamy:
closed because not really relevant according to default option `overrideWarAdditions`

src/it/test-dependencies-transitive-optional/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+            </dependency>
+            <dependency>
+                <groupId>org.jenkins-ci</groupId>
+                <artifactId>symbol-annotation</artifactId>
+                <version>1.20</version>
+            </dependency>
+            <dependency>
+                <groupId>ant</groupId>
+                <artifactId>ant</artifactId>
+                <!-- because old version have dependencies on xml-apis:xml-apis:2.6.2 which doesn't exists -->
```
olamy, line 138: too many bad poms in central...

src/it/override-test-dependencies-useUpperBounds/pom.xml
```diff
@@ -65,4 +65,31 @@
...
+                    </dependency>
+
+                    etc....
+                -->
+                <groupId>org.jenkins-ci.main</groupId>
+                <artifactId>jenkins-core</artifactId>
+                <version>${jenkins.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>jaxen</groupId>
```
olamy, line 88: because this pom has so bad dependencies :( https://repo.maven.apache.org/maven2/jaxen/jaxen/1.1.3/jaxen-1.1.3.pom

such 
```
<dependency>
<groupId>maven-plugins</groupId>
<artifactId>maven-cobertura-plugin</artifactId>
<version>1.3</version>
<type>plugin</type>
</dependency>
```

olamy:
Unfortunately even if this approach the problem is resolve dependencies `too deep` and try to get dependencies of bad old poms :( (bad metadata from the past)

src/main/java/org/jenkinsci/maven/plugins/hpi/JenkinsHpiDependencyCollector.java
```diff
@@ -0,0 +1,375 @@
...
+/**
+ * <p>
+ *     Project dependency raw dependency collector with Jenkins hpi filter need
+ * </p>
+ * <p>adaptation of some classes from
+ * https://github.com/apache/maven-dependency-tree/tree/master/src/main/java/org/apache/maven/shared/dependency/graph/internal
+ * </p>
+ */
+@Named
+public class JenkinsHpiDependencyCollector {
```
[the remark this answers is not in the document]
olamy: the only change is the removal of this `new OptionalDependencySelector()` this is our problem  https://github.com/apache/maven-dependency-tree/blob/fa81a70f7c6e3e37b77f57896156051273c0b2f3/src/main/java/org/apache/maven/shared/dependency/graph/internal/DefaultDependencyCollectorBuilder.java#L108 

but to keep using the mainstream code and not having to maintain a sort of forked class I have made this PR https://github.com/apache/maven-dependency-tree/pull/28
this will help us

src/it/override-test-dependencies-useUpperBounds/pom.xml
olamy: the one used here is 2.249 see https://github.com/jenkinsci/maven-hpi-plugin/blob/e8b202250a8f6512dbeaec97cdd83e595204dd68/src/it/override-test-dependencies-useUpperBounds/pom.xml#L16

src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
```diff
@@ -699,7 +696,12 @@ public Map<String, String> upperBounds(List<String> upperBoundsExcludes, String
 
                 // search for artifact with lowest hopCount
                 for (DependencyNodeHopCountPair hopPair : pairs.subList(1, pairs.size())) {
-                    if (hopPair.getHopCount() < resolvedPair.getHopCount()) {
+                    if (hopPair.getHopCount() < resolvedPair.getHopCount() ||
+                            // we can have dependencies with similar hop count but one is optional and the other not
+                            // Maven will use the non optional even if the version is higher
+                            (hopPair.getHopCount() == resolvedPair.getHopCount() &&
+                            resolvedPair.getNode().getArtifact().isOptional()) &&
+                            !hopPair.getNode().getArtifact().isOptional()) {
```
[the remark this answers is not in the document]
olamy: comment https://github.com/jenkinsci/maven-hpi-plugin/pull/391#discussion_r1016033257

src/main/java/org/jenkinsci/maven/plugins/hpi/JenkinsHpiDependencyCollector.java
```diff
@@ -0,0 +1,375 @@
...
+import org.eclipse.aether.util.graph.transformer.SimpleOptionalitySelector;
+import org.eclipse.aether.util.graph.visitor.TreeDependencyVisitor;
+import org.eclipse.aether.version.VersionConstraint;
+
+/**
+ * <p>
+ *     Project dependency raw dependency collector with Jenkins hpi filter need
+ * </p>
+ * <p>adaptation of some classes from
+ * https://github.com/apache/maven-dependency-tree/tree/master/src/main/java/org/apache/maven/shared/dependency/graph/internal
```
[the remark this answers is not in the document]
olamy: will not be needed with https://github.com/apache/maven-dependency-tree/pull/28

pom.xml
```diff
@@ -344,6 +354,9 @@
           <pomIncludes>
             <pomInclude>*/pom.xml</pomInclude>
           </pomIncludes>
+          <setupIncludes>
+            <setupInclude>setup_test-dependencies-transitive-optional-dep/pom.xml</setupInclude>
```
jglick, line 358: Does this really need to be done this way, or can you just do everything in one dir using a multimodule reactor?

src/it/test-dependencies-transitive-optional/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>copy-jenkins-war</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>copy</goal>
```
jglick, line 49: This does not look anything like a real plugin POM so I am struggling to follow what bug this IT purports to reproduce.

jglick (commented):
No description?

jtnord:
> as we are here in the case of some optional transitive dependencies.

but optional dependencies are not transitive so should not be present at all (so should not need updating).   
Something seems fishy @olamy ?

olamy:
> > as we are here in the case of some optional transitive dependencies.
> 
> but optional dependencies are not transitive so should not be present at all (so should not need updating). Something seems fishy @olamy ? (only direct (non-transitive) optional dependencies should be updated, or I am missing something?)

no of course optional dependencies are not transitive. But it looks there are some discrepancies between Jenkins plugins system (in PCT only?) and Maven dependencies (as you can read them in poms).

What I can see from the current case (issue from https://github.com/jenkinsci/bom/pull/3049)

before 
```
mina-sshd-api-core
+- ssh-credentials
|  \- trilead-api 
```

after
```
mina-sshd-api-core
+- ssh-credentials
|   @OptionalExtension(requirePlugins = {"trilead-api"})
|  \- trilead-api (optional)
```

As we can think, optional is not transitive when running `mina-sshd-api-core` tests in the PCT/bom context, the surefire classpath should not include `trilead-api` and it doesn't, so the CNFE.
**BUT** and it looks to be the real problem, while the extension is marked as `@OptionalExtension(requirePlugins = {"trilead-api"})` so you can you think as reading the poms this extension shouldn't be loaded as it's transitive optional.
But not Jenkins load because it can see the plugin but on the other side the classpath used by surefire doesn't contain the classes from trilead-api whereas it should.
So either we fix something in how Jenkins loads plugins in this case or we include the trilead-api in the classpath (i.e the transitive optional). (
Or as a workaround we keep adding non-needed dependencies in poms just to hide the root cause. (such https://github.com/jenkinsci/mina-sshd-api-plugin/pull/91 and there are probably few more)

jglick:
> loading of plugins/extensions […] should be consistent with the classpath calculation

In general, if you need to delve into this kind of detail, you should switch to `RealJenkinsRule`.

basil (commented):
Given that the essence of all the changes to `TestDependencyMojo` has been covered in other PRs, is there any reason to keep this PR open?

olamy:
good to see the changes have been integrated smoothly via other PRs

PR #393, opened 2022-11-12
Add a warning when using deprecated or insecure SCM URLs

basil:
A variety of plugins use deprecated (e.g. `git://``) or insecure (e.g., `http://`) SCM URLs, or URLs that cause problems for PCT. This PR adds a new (soft) warning that highlights such problems and provides a correction to the user. This shouldn't be very disruptive since I have already filed PRs to update the most egregious violations. I tested this on some of the same changes that I already filed PRs for to confirm that the suggestions match what I actually submitted in the PRs (before authoring this change).

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
@@ -2,6 +2,8 @@ package org.jenkinsci.maven.plugins.hpi;
 
 import hudson.util.VersionNumber;
 import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
+import java.util.Properties;
+import org.apache.maven.model.Scm;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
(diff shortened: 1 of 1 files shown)


MarkEWaite (approved):

src/main/java/org/jenkinsci/maven/plugins/hpi/ValidateMojo.java
```diff
@@ -15,6 +17,20 @@
...
+    private static final String SCM_GIT_GIT_URL_BAD = "scm:git:git://github.com/";
+    private static final String SCM_GIT_HTTP_URL_BAD = "scm:git:" + HTTP_GITHUB_COM;
+    private static final String SCM_GIT_SSH_URL_BAD = "scm:git:ssh://git@github.com/";
+
+    private static final String SCM_GIT_HTTPS_URL_GOOD = "scm:git:" + HTTPS_GITHUB_COM;
+    private static final String SCM_GIT_SSH_URL_GOOD = "scm:git:git@github.com:";
+
+    private static final String GIT_URLS_ARE_DEPRECATED = "git:// URLs are deprecated";
+    private static final String HTTP_URLS_ARE_INSECURE = "http:// URLs are insecure";
+    private static final String SSH_URLS_DO_NOT_WORK_WELL_WITH_PCT = "ssh:// URLs do not work well with PCT";
```
Vlatombe, line 32: I don't think it's really an argument. If they don't work well with pct as is, PCT should run with something like

```
[url "https://github.com/"]
	insteadOf = "git@github.com:"
	pushInsteadOf = "git@github.com:"
```

in its git configuration, guaranteeing that https is used in a homogeneous way even if the plugin has ssh configured.

Other variants can be used for other protocols.

PR #403, opened 2022-11-30
EOL `InitializeMojo`

basil:
Good riddance to this temporary hack.


Vlatombe:
Effectively reverted https://github.com/jenkinsci/maven-hpi-plugin/pull/323

PR #404, opened 2022-11-30
Clean up `TestRuntimeMojo`

basil:
Removing some no-longer-necessary Java 8 logic and maybe hoping to fix jenkinsci/bom#1551 by moving the copying of the patch module to an earlier phase? To test this I ran `mvn clean verify -Dtest=org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest`.

### DIFF
--- a/src/main/resources/META-INF/plexus/components.xml
+++ b/src/main/resources/META-INF/plexus/components.xml
@@ -18,7 +18,8 @@
               <process-test-resources>org.apache.maven.plugins:maven-resources-plugin:testResources</process-test-resources>
               <generate-test-sources>org.jenkins-ci.tools:maven-hpi-plugin:insert-test</generate-test-sources>
               <test-compile>org.apache.maven.plugins:maven-compiler-plugin:testCompile,org.jenkins-ci.tools:maven-hpi-plugin:test-hpl,org.jenkins-ci.tools:maven-hpi-plugin:resolve-test-dependencies</test-compile>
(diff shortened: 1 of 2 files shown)


src/main/resources/META-INF/plexus/components.xml
```diff
@@ -18,7 +18,8 @@
               <process-test-resources>org.apache.maven.plugins:maven-resources-plugin:testResources</process-test-resources>
               <generate-test-sources>org.jenkins-ci.tools:maven-hpi-plugin:insert-test</generate-test-sources>
               <test-compile>org.apache.maven.plugins:maven-compiler-plugin:testCompile,org.jenkins-ci.tools:maven-hpi-plugin:test-hpl,org.jenkins-ci.tools:maven-hpi-plugin:resolve-test-dependencies</test-compile>
-              <test>org.jenkins-ci.tools:maven-hpi-plugin:test-runtime,org.apache.maven.plugins:maven-surefire-plugin:test</test>
```
basil, line 21: Guessing that maybe the order of execution is undefined when these are in the same lifecycle phase?

basil (commented):
Expediting this but feel free to review and I will address any comments later

basil:
Based on https://github.com/jenkinsci/bom/pull/1615/checks?check_run_id=9822951672 this didn't seem to help with the `java.lang.NoClassDefFoundError: org/netbeans/insane/hook/MakeAccessible` error, but at least it didn't hurt, so that's good

PR #409, opened 2022-12-07
Upgrade Jetty from 9.4.49.v20220914 to 10.0.12

basil:
Now that the HPI plugin requires Java 11, we can move to a Jetty 10 based implementation To test this I ran `mvn hpi:run` before and after this change, ensuring that Jenkins came up successfully in both cases and could be reloaded by pressing the enter key.


.github/dependabot.yml
```diff
@@ -1,15 +1,16 @@
 # https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
+---
```
basil, line 2: Making `yamllint(1)` happy.

README.md
```diff
@@ -36,10 +36,3 @@ groovy -e "basedir='$(pwd)/target/its/parent-4x'; evaluate new File('src/it/pare
 ```
 
 Also make sure `project.parent.version` is the latest in every integration test except `src/it/parent-4-40/pom.xml`.
-
-## Updating Jetty
-`hpi:run` mojo is a variant of `jetty:run` mojo, and because of the way plugin descriptor is generated, this module copies some code from Jetty Maven plugin, specifically `AbstractJettyMojo.java` and `ConsoleScanner.java`.
-
-To keep upstream tracking easier, pristine copies of these files are copied into `incoming-x.y` branch, then package renamed. This version specific incoming branch is then "theirs" merged into the `incoming` branch, which acts as the upstream tracking branch.
-
-This branch is then merged into `master` via `git merge -X ignore-space-at-eol incoming`. See diff between `incoming` and `master` on these files to see the exact local patches.
```
basil, line 45: No longer relevant, as we no longer copy code from Jetty Maven plugin.

pom.xml
```diff
@@ -161,6 +161,12 @@
       <groupId>org.apache.maven.plugins</groupId>
       <artifactId>maven-dependency-plugin</artifactId>
       <version>3.4.0</version>
+      <exclusions>
+        <exclusion>
+          <groupId>asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+      </exclusions>
```
basil, line 169: Was pulling in an ancient version of ASM 3, which is not compatible with Jetty Maven plugin.

pom.xml
```diff
@@ -202,6 +208,12 @@
       <groupId>org.kohsuke.stapler</groupId>
       <artifactId>stapler-groovy</artifactId>
       <version>1756.vec9071748061</version>
+      <exclusions>
+        <exclusion>
+          <groupId>jakarta.annotation</groupId>
+          <artifactId>jakarta.annotation-api</artifactId>
+        </exclusion>
+      </exclusions>
```
basil, line 216: Was pulling in a Jakarta import version of this API which conflicts with that of Jetty.

pom.xml
```diff
@@ -313,9 +325,9 @@
                 <requireUpperBoundDeps>
                   <excludes combine.children="append">
                     <exclude>commons-lang:commons-lang</exclude>
+                    <exclude>org.apache.commons:commons-collections4</exclude>
                     <exclude>org.codehaus.plexus:plexus-archiver</exclude>
                     <exclude>org.codehaus.plexus:plexus-container-default</exclude>
-                    <exclude>org.ow2.asm:asm</exclude>
```
basil, line 318: Miscellaneous cleanup: realized that this was unnecessary while updating this list.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -161,7 +163,7 @@ public class RunMojo extends AbstractJettyMojo {
      *
      * If connectors are configured in the Mojo, that'll take precedence.
      */
-    @Parameter(property = "port")
+    @Parameter(property = "port", defaultValue = "8080")
     protected int defaultPort;
```
basil, line 167: Jetty's behavior changed to give us a random port every time. To preserve the existing behavior (which defaults to port 8080) I had to add this. In the future, we could enhance HPI plugin to work with Jetty's new random port behavior, but the logic to print the URL would need to be adjusted. Since the goal of this PR was to do the minimum amount of change necessary to complete the upgrade, I did not tackle that in this PR.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -183,7 +185,10 @@ public class RunMojo extends AbstractJettyMojo {
     /**
      * If true, the context will be restarted after a line feed on
      * the input console. Enabled by default.
+     *
+     * @deprecated use {@link JettyRunWarMojo#scan}
      */
+    @Deprecated
     @Parameter(property = "jetty.consoleForceReload", defaultValue = "true")
     protected boolean consoleForceReload;
```
basil, line 193: The superclass now has a more fine-grained `scan` parameter to control reloading via console or by scanning files for changes.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -225,14 +230,6 @@ public class RunMojo extends AbstractJettyMojo {
...
     protected boolean pluginFirstClassLoader = false;
 
-    /**
-     * List of additional System properties to set
-     *
-     * @since 1.85
-     */
-    @Parameter
-    private Map<String, String> systemProperties;
-
```
basil, line 235: This is now present in the superclass so there is no need to maintain our own shadow field.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -277,7 +274,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 getLog().warn("Please use `webApp/contextPath` configuration parameter in place of the deprecated `contextPath` parameter");
                 if (webApp == null) {
                     try {
-                        webApp = new JettyWebAppContext();
+                        webApp = new MavenWebAppContext();
```
basil, line 277: This class was renamed upstream.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -312,14 +309,6 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
         setSystemPropertyIfEmpty("jenkins.moduleRoot", basedir.getAbsolutePath());
 
-        if (systemProperties != null && !systemProperties.isEmpty()) {
-            for (Map.Entry<String,String> entry : systemProperties.entrySet()) {
-                if (entry.getKey() != null && entry.getValue()!=null) {
-                    System.setProperty( entry.getKey(), entry.getValue() );
-                }
-            }
-        }
-
```
basil, line 322: This is now done by the superclass so there is no need to do it ourselves.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -512,17 +501,22 @@ private void generateHpl() throws MojoExecutionException {
...
+            webApp.setTempDirectory(tmp.toFile());
+        }
+        File extractedWebAppDir = new File(webApp.getTempDirectory(), "webapp");
         if (isExtractedWebAppDirStale(extractedWebAppDir, webAppFile)) {
             FileUtils.deleteDirectory(extractedWebAppDir);
         }
-        
-        super.configureWebApplication();
         getWebAppConfig().setWar(webAppFile.getCanonicalPath());
+        super.configureWebApp();
```
basil, line 519: This has to be done after setting the WAR rather than before in order to avoid tripping an assertion.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -540,6 +534,7 @@ public void configureWebApplication() throws Exception {
         userStore.addUser("bob", new Password("bob"), new String[] {"user", "male"});
         userStore.addUser("charlie", new Password("charlie"), new String[] {"user", "male"});
         getWebAppConfig().getSecurityHandler().setLoginService(hashLoginService);
+        finishConfigurationBeforeStart();
```
basil, line 537: The super class no longer provides two distinct hooks here, but to keep the diff small I retained the existing split between two methods and made the first vector into the second.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -605,93 +600,30 @@ private String loadVersion(InputStream is) throws IOException {
     }
 
     @Override
-    public void configureScanner() throws MojoExecutionException {
+    public void startScanner() throws Exception {
```
basil, line 603: Adapting the hack to changes in the superclass's calling conventions. See the comment for explanation.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -605,93 +600,30 @@ private String loadVersion(InputStream is) throws IOException {
...
+        // this can only be done after MavenServerConnector.doStart() is called, which happens in AbstractWebAppMojo.startJetty()
+        // and this startScanner method is one of the few places that are run afterward.
+        HttpConfiguration hc = httpConnector.getConnectionFactory(HttpConnectionFactory.class).getHttpConfiguration();
+        hc.setHttpCompliance(HttpCompliance.RFC7230);
+        hc.setUriCompliance(UriCompliance.LEGACY);
+        // Use a bigger buffer, as Stapler traces can get pretty large on deeply nested URLs.
+        hc.setResponseHeaderSize(12 * 1024);
+
+        super.startScanner();
     }
```
basil, line 614: Was copypasta from the superclass with no customizations, except for the call to `generateHpl()` which was dead code anyway.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -605,93 +600,30 @@ private String loadVersion(InputStream is) throws IOException {
...
-            consoleScanner = new ConsoleScanner(this);
-            consoleScanner.start();
+    protected boolean isPackagingSupported() {
+        if (!supportedPackagings.contains("hpi")) {
+            List<String> newSupportedPackagings = new ArrayList<>(supportedPackagings);
+            newSupportedPackagings.add("hpi");
+            supportedPackagings = List.copyOf(newSupportedPackagings);
         }
+        return super.isPackagingSupported();
     }
```
basil, line 624: I considered forcing consumers to add

```xml
              <supportedPackagings>
                <supportedPackaging>hpi</supportedPackaging>
              </supportedPackagings>
```

but felt it would be smoother if the default included `hpi`. I didn't remove `war` but I don't feel strongly about this.

.github/dependabot.yml
jglick: Or just delete the comment since the link is dead anyway!

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
jglick: I think we would prefer to keep the static port anyway, since this is for interactive use and having a stable root URL across runs is desirable.

jglick (approved):
Thanks!

.github/dependabot.yml
basil: Pre-existing issue

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
basil: Yeah, I felt the same way after having a page open in my browser and wanting to refresh it and get an updated version without having to change the port.

PR #411, opened 2022-12-08
Remove no longer needed workarounds for Jenkins modules

basil:


daniel-beck:
Per https://github.com/jenkinsci/jep/blob/master/jep/230/README.adoc,

> The technical ability to package JARs as Jenkins modules is retained

so this isn't motived by JEP-230. Could you add a motivation to your PR description?

basil:
Whatever this code is trying to do, it could not possibly be correct because it is doing it based on the assumption that something with the `jenkins-module` package name is a special case module, not a plugin. These have all been converted to plugins now but retain that package name, so the regular logic for plugins should be applying to them. I felt it simplest just to delete this unneeded code now that there are no modules left.

basil:
Oh I think I misread this: I thought it was referring to the `org.jenkins-ci.modules` group ID but it is actually referring to the `packaging` type which nothing is actually using anymore. So I guess it is harmless to keep it.

PR #415, opened 2022-12-23
hpi:run uses wrong jpi/hpl from .jenkins-hpl-map

jgreffe:
https://issues.jenkins.io/browse/JENKINS-70329

Check if the resolved `the.hpl` from `/Users/xxx/.jenkins-hpl-map` are contained within the current maven root project.
If not, logs a WARN instead of an INFO.

### DIFF
--- a/src/it/JENKINS-70329-dir-one/pom.xml
+++ b/src/it/JENKINS-70329-dir-one/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>4.53</version>
+    <relativePath />
+  </parent>
+
+  <groupId>org.jenkins-ci.tools.hpi.its</groupId>
+  <artifactId>duplicate-bundle-it</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <packaging>pom</packaging>
+
+  <properties>
+    <jenkins.version>2.361.4</jenkins.version>
+    <hpi-plugin.version>@project.version@</hpi-plugin.version>
+  </properties>
+  
+  <modules>
+    <module>sub-module1</module>
+    <module>sub-module2</module>
+    <module>sub-module3</module>
+    <module>sub-module4</module>
+    <module>sub-module5</module>
+    <module>sub-module6</module>
+    <module>sub-module7</module>
+    <module>sub-module8</module>
+    <module>sub-module9</module>
+    <module>sub-module10</module>
+  </modules>
+</project>
--- a/src/it/JENKINS-70329-dir-two/invoker.properties
+++ b/src/it/JENKINS-70329-dir-two/invoker.properties
@@ -1,0 +1,20 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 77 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -43,6 +43,7 @@
 import org.apache.maven.project.ProjectDependenciesResolver;
 import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolver;
 import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolverException;
+import org.apache.tools.ant.Project;
```
jtnord, line 46: unsued and wrong import?

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
```diff
@@ -418,6 +421,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
 
+    private File getRootBasedir(MavenProject mavenProject) {
+        if (mavenProject.getParent() == null || 
+            mavenProject.getParent().getBasedir() == null ||
+            !mavenProject.getBasedir().getAbsolutePath()
+                         .contains(mavenProject.getParent().getBasedir().getAbsolutePath())) {
+            return mavenProject.getBasedir();
+        }
+        return getRootBasedir(mavenProject.getParent());
+    }
```
jtnord, line 432: this does not appear to be correct. There is no guarantee that a parent in of any submodule in a build is in the reactor, or parent structure.  

For more reading check the [difference between an aggregator project and a parent](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance_vs_Project_Aggregation).  It is possible to be both an aggregator and a parent, but it is not required.

src/it/JENKINS-70329-dir-one/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+    <module>sub-module1</module>
+    <module>sub-module2</module>
+    <module>sub-module3</module>
+    <module>sub-module4</module>
+    <module>sub-module5</module>
+    <module>sub-module6</module>
+    <module>sub-module7</module>
+    <module>sub-module8</module>
+    <module>sub-module9</module>
+    <module>sub-module10</module>
```
jtnord, line 35: do you really need 10 modules to validate this?

src/it/JENKINS-70329-dir-two/invoker.properties
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+invoker.goals=-ntp clean install hpi:run -Djetty.skip=true -DskipTests
```
jtnord, line 20: this obviously works - but how come?  `hpi:run` would normally run forever - does it terminate as there is no console?

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
jtnord: you could look at the current maven session and what modules are in it. 
or you could possibly use the `maven.multiModuleProjectDirectory` property.

src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
jgreffe: Indeed

src/it/JENKINS-70329-dir-two/invoker.properties
jgreffe: Property `-Djetty.skip=true` avoids the jetty server to run, we don't start any infinite process.

src/it/JENKINS-70329-dir-one/pom.xml
jgreffe: It's an arbitrary number of modules, could be less, but we may have it working by chance even without the fix.

jglick (changes requested):
Not the right fix I think. Fails to diagnose the essential problem: that the `*.properties` file is randomly sorted, rather than chronologically.

Probably it would suffice to just invert the map, so that it is from id to file, so that `mvn install` on a new checkout would overwrite previous entries of the same GAV. I cannot recall why it was in the current order to begin with.

jglick:
> I cannot recall why it was in the current order to begin with.

One reason I remember now: in the normal case that the snapshot version changes (in the case of core or a plugin using `maven-release-plugin` rather than CD), you would otherwise get an ever-growing list of entries with the same `groupId` / `artifactId` but different `versions`, with no garbage collection. The current scheme at least ensures that after a version change (in the same checkout directory), the stale entry is overwritten.

An alternative scheme which would address [JENKINS-70329](https://issues.jenkins.io/browse/JENKINS-70329) simply while avoiding an ever-growing file might be to write out directory / GAV pairs in reverse chronological order (implies using some format other than `Properties`, which discards order; see also #428) and discard anything beyond the most recent 100 or 1000 or whatever.

jgreffe:
Obsolete

PR #420, opened 2023-01-09
do not try and unpack all our dependencies looking for a mojo

jtnord:
The plugins Mojos are not defined in dependencies, so do not try and unpack them looking for them

This fixes a slew of warnings when building the plugin on windows: 

```
[WARNING] Archive entry
'org/apache/maven/plugins/dependency/fromDependencies/AbstractFromDependenciesMojo.java' and existing file
'C:\workarea\source\github\jenkinsci\maven-hpi-plugin\target\maven-plugin-plugin-sources\org.apache.maven.plugins\maven-dependency-plugin\3.4.0\sources\org\apache\maven\plugins\dependency\fromDependencies\AbstractFromDependenciesMojo.java' names differ only by case. This may lead to an unexpected outcome on case-insensitive filesystems.
```

This fixes the noise spam.  really the unarchiver that is erroneously warning (the names do match just `/` is a `\`) but I could not quickly find the source of this, and this fix means we can  see any other warnings.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -376,8 +376,8 @@
         <version>${maven-plugin-tools.version}</version>
         <configuration>
           <goalPrefix>hpi</goalPrefix>
-          <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
-          <!-- see MNG-5346 -->
+          <extractors>java-annotations</extractors>
+          <mojoDependencies></mojoDependencies>
         </configuration>
         <executions>
           <execution>
--- a/pom.xml
+++ b/pom.xml
@@ -377,7 +377,7 @@
         <configuration>
           <goalPrefix>hpi</goalPrefix>
           <extractors>java-annotations</extractors>
-          <mojoDependencies></mojoDependencies>
+          <mojoDependencies>org.eclipse.jetty:jetty-maven-plugin</mojoDependencies>
         </configuration>
         <executions>
           <execution>


pom.xml
```diff
@@ -376,8 +376,8 @@
         <version>${maven-plugin-tools.version}</version>
         <configuration>
           <goalPrefix>hpi</goalPrefix>
-          <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
```
jtnord, line 379: no longer required so removed.

pom.xml
```diff
@@ -376,8 +376,8 @@
...
-          <!-- see MNG-5346 -->
+          <extractors>java-annotations</extractors>
```
jtnord, line 379: we are only using java annotations, not javadoc, so no need to look for javadoc comments, or bsh or the other deprecated types.

pom.xml
```diff
@@ -376,8 +376,8 @@
...
+          <mojoDependencies></mojoDependencies>
```
jtnord, line 380: as all our mojos are not in dependencies tell maven not too look in any of our dependencies (some of this are plugins which is interesting!)

by default this list is `null` which means check everything, and empty list is check no dependencies

jtnord: https://maven.apache.org/plugin-tools/maven-plugin-plugin/descriptor-mojo.html#mojoDependencies

basil: > as all our mojos are not in dependencies

Actually one of them is: `jetty-maven-plugin`, needed for `RunMojo` (and why your tests were failing).

basil (approved):
In the future please do not request a review from me until your tests are passing.

jtnord:
> In the future please do not request a review from me until your tests are passing.

Apologies, I had unrequested your review as soon as I saw the failure in CI 
image

it was tested locally (mvn verify) locally but I had not realised that the project does not actually run ITs - I dislike setups where running tests locally does not run the actual tests as surprises like this happen)

jtnord:
this PR was not complete - 51 warnings still remain due to the unpacking, but are now limited to the Jetty maven plugin.

PR #436, opened 2023-02-17
Add additional plugin manifiest entries

jtnord:
Adds the following Manifest entries

Artifact-Id : whilst there was Short-Name that used artifactId it could
	be changed (however unlikely) and it is best to have something that won't ever change.
	Could have been obtained from the pom.xml or pom.properties - but
	that requires you to go looking
Plugin-Scm-Connection : currently the PCT is forcing plugins to specify
	this in mulit module projects which is convoluted.  Also moving here could
	simplify in the future by removing the need to parse any poms.
Plugin-Scm-Git-Hash :  put the Git SHA used for the HEAD of the plugin
	build. Allows the (best effort) recording of a commit used for a SNAPSHOT build.
Plugin-Scm-Git-Module-Path : the path of the project relative to the git
	project.

Regardless of any future PCT work to use these additions `Plugin-Scm-Git-Hash` is useful when looking at a deployed SNAPSHOT today and the others do no harm.


src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -827,31 +827,36 @@ private static void copyFilteredFile(File from, File to, String encoding, Filter
     /**
      * If the project is on Git, figure out Git SHA1.
      *
-     * @return null if no git repository is found
-     */
-    public String getGitHeadSha1() {
-        // we want to allow the plugin that's not sitting at the root (such as findbugs plugin),
-        // but we don't want to go up too far and pick up unrelated repository.
-        File git = new File(project.getBasedir(), ".git");
```
jtnord, line 835: this was questionable to begin with.
it implies that
1. you can only ever have a 1 level structure
2. it would be valid and useful to checkout a non git project into a clone of a git projects directory structure (but only at 2 levels or lower!)

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -827,31 +827,36 @@ private static void copyFilteredFile(File from, File to, String encoding, Filter
...
+            return null;    // git repository present, but without commits
+        } 
+        if (shortHash) {
+            return v.substring(0,8);
+        }
+        return v;
+    }
+    
+    protected String getGitRevParseOutput(String revparseArg) {
+        if (!project.getScm().getConnection().startsWith("scm:git")) {
```
jtnord, line 848: rather than looking for a `.git` folder in the hierachy we now just ask the project if its connection starts with `scm:git`

This does not need to be set for every plugin in a reactor, as the model we have contains inherited data.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -124,6 +124,8 @@ protected void setAttributes(Manifest.ExistingSection mainSection) throws MojoEx
         }
 
         mainSection.addAttributeAndCheck(new Manifest.Attribute("Group-Id",project.getGroupId()));
+        mainSection.addAttributeAndCheck(new Manifest.Attribute("Artifact-Id",project.getArtifactId()));
```
jtnord, line 127: There is no harm in adding this value explicitly.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -213,7 +215,11 @@ protected void setAttributes(Manifest.ExistingSection mainSection) throws MojoEx
         addLicenseAttributesForManifest(mainSection);
         addPropertyAttributeIfNotNull(mainSection, "Plugin-ChangelogUrl", "hpi.pluginChangelogUrl");
         addPropertyAttributeIfNotNull(mainSection, "Plugin-LogoUrl", "hpi.pluginLogoUrl");
+
         addAttributeIfNotNull(mainSection, "Plugin-ScmUrl", getScmUrl());
+        addAttributeIfNotNull(mainSection, "Plugin-Scm-Connection", getSCMConnection());
```
jtnord, line 220: so that it could be obtained in the future without projects having to pollute their own pom.

jenkinsci/plugin-compat-tester#453

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -213,7 +215,11 @@ protected void setAttributes(Manifest.ExistingSection mainSection) throws MojoEx
...
+        addAttributeIfNotNull(mainSection, "Plugin-Scm-Git-Hash", getGitHeadSha1(false));
```
jtnord, line 221: useful for tracking SNAPSHOT deployments.  

Possibly helpful for testing SNAPSHOT builds in the PCT.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -213,7 +215,11 @@ protected void setAttributes(Manifest.ExistingSection mainSection) throws MojoEx
...
+        addAttributeIfNotNull(mainSection, "Plugin-Scm-Git-Module-Path", getGitModulePath());
```
jtnord, line 222: does no harm as is.

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractJenkinsManifestMojo.java
```diff
@@ -292,6 +298,37 @@ private String getScmUrl() {
...
+            String b = project.getBasedir().getCanonicalPath();
+            // no need to be defensive - the project must be in the git directory for git-revparse to return non null
+            String path = b.substring(f.length());
+            if (path.startsWith(File.separator)) {
+                path = path.substring(1); // for a single module project this will be the empty string
+            }
+            // normalize to Unix style
+            return path.replace(File.separatorChar, '/');
+        } catch (IOException e) {
+            getLog().warn("failed to obtain projects relative location in the Git repository", e);
```
jtnord, line 327: goes against the `fine` level if `git rev-parse HEAD` fails in the existing code - but really this should not happen - we are a git project.

basil (approved):
I simplified this considerably.

PR #451, opened 2023-03-15
fix filtering of web resources using system properties and escaping

jtnord:
* Filtering `webResources` now consults `System.getProperties()`  (again?)
* Filtering is now escapable so you can write `\${project.version}` and not have it replaced by the project version. (this matches Mavens' default behaviour

Replaces code from plexus-util with maven-filtering.

The code still directly uses some plexus-util so the dependency is not removed but the comment was no longer correct so that was removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,11 @@
     </dependency>
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
+      <artifactId>maven-filtering</artifactId>
+      <version>3.3.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-dependency-tree</artifactId>
       <version>3.2.1</version>
     </dependency>
@@ -212,7 +217,6 @@
     <dependency>
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-utils</artifactId>
-      <!-- generated help mojo has a dependency on plexus utils -->
       <version>3.5.1</version>
     </dependency>
     <dependency>
--- a/src/it/warResources/src/main/webResource-filtering/help-globalConfig.html
+++ b/src/it/warResources/src/main/webResource-filtering/help-globalConfig.html
@@ -7,12 +7,12 @@
   </p>
   <p>
     You can also filter using interpolation like: 
-    \${prop1\} ${prop1}
+    \${prop1} ${prop1}
     or 
     \@prop2\@ @prop2@
     and these will be replaced at build time
     it is even possible to use system properties
-    \@java.class.version\@ ${java.class.version}
+    \@java.vendor.url\@ ${java.vendor.url}
   </p>
 
 </div>
--- a/src/it/warResources/verify.groovy
+++ b/src/it/warResources/verify.groovy
@@ -17,11 +17,9 @@
  * under the License.
  */
 globalConfigHelp  = new File(basedir, 'target/war-resources-it/help-globalConfig.html').text;
-// code uses plexus not maven so escapes are not supported.
-assert globalConfigHelp.contains("\\\${prop1\\} hello");
-assert globalConfigHelp.contains("\\@prop2\\@ goodbye");
-// code claims to add System properties but this fails
-// assert globalConfigHelp.contains("\\@java.class.version\\@ 55.0");
+assert globalConfigHelp.contains('${prop1} hello');
+assert globalConfigHelp.contains('@prop2@ goodbye');
+assert globalConfigHelp.contains('@java.vendor.url@ https://');
 
 
 return true;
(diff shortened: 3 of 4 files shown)


pom.xml
```diff
@@ -212,7 +217,6 @@
     <dependency>
       <groupId>org.codehaus.plexus</groupId>
       <artifactId>plexus-utils</artifactId>
-      <!-- generated help mojo has a dependency on plexus utils -->
```
jtnord, line 215: the generated file did not contain any plexus-util imports

src/it/warResources/verify.groovy
```diff
@@ -20,8 +20,7 @@ globalConfigHelp  = new File(basedir, 'target/war-resources-it/help-globalConfig
 // code uses plexus not maven so escapes are not supported.
 assert globalConfigHelp.contains("\\\${prop1\\} hello");
 assert globalConfigHelp.contains("\\@prop2\\@ goodbye");
-// code claims to add System properties but this fails
-// assert globalConfigHelp.contains("\\@java.class.version\\@ 55.0");
+assert globalConfigHelp.contains("\\@java.vendor.url\\@ https://");
```
jtnord, line 23: System properties now work (which was expected before from the code - but did not actually work)

src/it/warResources/verify.groovy
```diff
@@ -20,8 +20,7 @@ globalConfigHelp  = new File(basedir, 'target/war-resources-it/help-globalConfig
 // code uses plexus not maven so escapes are not supported.
```
jtnord, line 20: this comment is now incorrect.

I have not enabled escapes - but it is trivial to do so.  this may be more of a breaking change, however I found only a handful of projects (mostly legacy) that used `webResources` and none used filtering so I could make that change here too (it is only an extra line of code)

src/it/warResources/src/main/webResource-filtering/help-globalConfig.html
```diff
@@ -7,12 +7,12 @@
...
   <p>
     You can also filter using interpolation like: 
-    \${prop1\} ${prop1}
+    \${prop1} ${prop1}
     or 
     \@prop2\@ @prop2@
     and these will be replaced at build time
     it is even possible to use system properties
-    \@java.class.version\@ ${java.class.version}
+    \@java.vendor.url\@ ${java.vendor.url}
```
jtnord, line 15: class version is of the JVM not the compiled classes - so this would on java17 etc.  replaced with vendor.rul which is very jvm specific - but it is 2023 so if this is not a https:// url then someone should fix the JVM...

src/main/java/org/jenkinsci/maven/plugins/hpi/AbstractHpiMojo.java
```diff
@@ -403,28 +383,16 @@ private Properties getBuildFilterProperties()
...
-                }
-            }
-        }
+     * @throws MavenFilteringException if an error occurred while copying webResources
+     */
+    public void copyResourcesWithFiltering(List<Resource> resources, File webappDirectory) throws MavenFilteringException {
+        MavenResourcesExecution mavenResourcesExecution = 
+                new MavenResourcesExecution(resources, webappDirectory, project, StandardCharsets.UTF_8.name(), filters,
+                                              Collections.emptyList(), session);
+        mavenResourcesExecution.setEscapeString("\\");
```
jtnord, line 394: this means we can escape things - which is what you would get with regular maven filtering.

technically this is a breaking change - but looking through projects of the 5 I found adding `webResources` non used filtering anyway... so making this compatible would appear to be overkill

An argument could be made therefor to remove this code entirely - but as the `webResources` are of type [`Resource`](https://maven.apache.org/ref/3.9.0/maven-model/apidocs/org/apache/maven/model/Resource.html) it would mean we would take options that would be discarded.

src/it/warResources/verify.groovy
```diff
@@ -17,11 +17,9 @@
...
  */
 globalConfigHelp  = new File(basedir, 'target/war-resources-it/help-globalConfig.html').text;
-// code uses plexus not maven so escapes are not supported.
-assert globalConfigHelp.contains("\\\${prop1\\} hello");
-assert globalConfigHelp.contains("\\@prop2\\@ goodbye");
-// code claims to add System properties but this fails
-// assert globalConfigHelp.contains("\\@java.class.version\\@ 55.0");
+assert globalConfigHelp.contains('${prop1} hello');
+assert globalConfigHelp.contains('@prop2@ goodbye');
+assert globalConfigHelp.contains('@java.vendor.url@ https://');
```
jtnord, line 22: showing a System property is now resolved.

src/it/warResources/verify.groovy
```diff
@@ -17,11 +17,9 @@
  * under the License.
  */
 globalConfigHelp  = new File(basedir, 'target/war-resources-it/help-globalConfig.html').text;
-// code uses plexus not maven so escapes are not supported.
-assert globalConfigHelp.contains("\\\${prop1\\} hello");
-assert globalConfigHelp.contains("\\@prop2\\@ goodbye");
-// code claims to add System properties but this fails
-// assert globalConfigHelp.contains("\\@java.class.version\\@ 55.0");
+assert globalConfigHelp.contains('${prop1} hello');
```
jtnord, line 20: the escapes are now working - so they get replaced with the literal (in other words the escape string is correctly removed.

basil (approved):

olamy (approved):

PR #454, opened 2023-03-26
Resolve test dependencies in parallel

basil:
I have long been annoyed at the fact that `resolve-test-dependencies` downloads one dependency at a time. This is made worse by my slow Internet connection. This PR makes it work in parallel, the same way regular Maven dependency resolution works.

Unfortunately `maven-artifact-transfer` (the nice high-level API we were using) does not support this use case. It was a very thin layer over Aether, anyway. Here we use Aether directly, which can do what we want. There is precedent for invoking Aether directly in other Maven plugins, such as `maven-enforcer-plugin`.

We do at most two "passes": one regular pass, and one (possibly skipped) fallback pass to retain the functionality of JENKINS-62658. Each pass resolves multiple artifacts in parallel for maximum performance.

This entails a bit of code duplication, but I think the performance improvement is worth it.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
@@ -1,12 +1,10 @@
 package org.jenkinsci.maven.plugins.hpi;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.BufferedWriter;
 import java.io.File;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStreamWriter;
-import java.io.Writer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


basil:
In `jenkinsci/bom`, before:

```
[2023-03-26T04:40:45.726Z] [INFO] --- hpi:3.39:resolve-test-dependencies (default-resolve-test-dependencies) @ sample ---
[2023-03-26T04:41:31.413Z] [INFO] 
```

After:

```
[2023-03-26T19:00:54.353Z] [INFO] --- hpi:3.41-rc1397.a_3fb_b_d24b_e6b_:resolve-test-dependencies (default-resolve-test-dependencies) @ sample ---
[2023-03-26T19:00:57.456Z] [INFO] 
```

46 seconds to 3 seconds (> 1000% performance improvement)

I also tested this manually in `text-finder` by adding the latest version of all the AWS SDK plugins to my POM in test scope. Before this PR they were painfully downloading one-by-one. After this PR they were downloading in parallel.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
```diff
@@ -390,6 +374,105 @@ public void execute() throws MojoExecutionException {
...
+            }
+
+            String artifactId;
+            try {
+                artifactId = mavenArtifact.getActualArtifactId();
+            } catch (IOException e) {
+                throw new MojoExecutionException("Failed to resolve " + mavenArtifact.getId(), e);
+            }
+            if (artifactId == null) {
+                getLog().debug("Skipping " + artifactId + " with classifier " + mavenArtifact.getClassifier());
```
MarkEWaite, line 403: Since `artifactId` is null, I think this will be a constant string.  Optional to report the constant string directly rather than having Java convert the null to the string "null".  This is not a blocking comment, since the code is well behaved as written.
```suggestion
                getLog().debug("Skipping null artifactID with classifier " + mavenArtifact.getClassifier());
```

MarkEWaite (approved):
Approved.  Comment for the debug logging is optional

src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
basil: Sure, this code was there before this PR as well:

```
                if (artifactId == null) {
                    getLog().debug("Skipping " + artifactId + " with classifier " + a.getClassifier());
                    continue;
                }
```

To be honest, I think it is pointless and could just be deleted, but I was trying to touch the original code as little as possible. But I accepted your suggestion because it is an improvement over the status quo.

PR #461, opened 2023-03-28
Remove `overrideWarAdditions` option

basil:
`jenkinsci/bom` is using `-DoverrideWarAdditions`, while CloudBees is not. This PR removes the most controversial part of `-DoverrideWarAdditions` (adding the entries to test scope) while retaining the part of it that is needed for `jenkinsci/bom` (adding the bundled plugins that are not part of the original dependency tree to the dependency management section), now the default behavior. Hopefully this will not cause any problems for CloudBees so that we can unify the two implementations.

I plan to test this in `jenkinsci/bom` and would request someone also test this on the CloudBees side.


jtnord (approved):

basil:
Ready to go with this @jtnord? I saw that the internal test of https://github.com/jenkinsci/plugin-compat-tester/pull/505 passed.

jtnord (approved):

PR #478, opened 2023-05-03
Handle SNAPSHOT versions in with `overrideWar`

jtnord:
if the war scanning encounters a SNAPSHOT build of a plugin then then manifest file will contain a `Plugin-Version` entry like `1.2.3-4-SNAPSHOT (private abcd1234-username)`.

This causes a failure when updating the dependencies (`resolve-test-dependencies`) as the plugin will try and use that directly without stripping off the extra parts.

This strips off any extra parts should they be present so that the version number matches a valid maven version number and also emits a warning to the log about non repeatable builds if any snapshots are encountered.

### Testing done

manually tested with a war containing a snapshot using the command

`mvn.cmd -B -V -e -DoverrideWar=C:\workarea\myWarWithSnapshot.war -Djenkins.version=2.387.3-cb-2 -DuseUpperBounds=true org.jenkins-ci.tools:maven-hpi-plugin:3.44-20230503.112206-2:resolve-test-dependencies`

mojo no longer fails and the log is present in the output

```
...
[INFO] --- hpi:3.44-SNAPSHOT:resolve-test-dependencies (default-cli) @ operations-center-server ---
[WARNING] War contains a SNAPSHOT of com.cloudbees.operations-center.common:operations-center-context build will not be fully repeatable
[INFO] Scanned contents of C:\workarea\myWarWithSnapshot.wa with 155 bundled plugins
[INFO] Updating direct dependency io.jenkins.plugins:okhttp-api from 4.9.3-108.v0feda04578cf to 4.10.0-132.v7a_7b_91cef39c
...
[INFO] Updating direct dependency com.cloudbees.operations-center.common:operations-center-context from 2.387.0.3-JTN-SNAPSHOT to 2.387.0.3-JTN-SNAPSHOT
...
```

IT in #480

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
@@ -552,8 +552,8 @@ public class TestDependencyMojo extends AbstractHpiMojo {
                         // handle any extra info in snapshots e.g. " (private-abcd1234-username)"
                         version = version.replaceFirst(" [(].+[)]$", "");
                         if (version.endsWith("-SNAPSHOT")) {
-                            getLog().warn("War contains a SNAPSHOT of " + groupId + ":" + artifactId
-                                    + " build will not be fully repeatable");
+                            getLog().warn("WAR contains a SNAPSHOT of " + groupId + ":" + artifactId
+                                    + "; build will not be fully repeatable");
                         }
                         String key = String.format("%s:%s", groupId, artifactId);
                         String self = String.format("%s:%s", project.getGroupId(), project.getArtifactId());
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
@@ -146,7 +146,7 @@ public class TestDependencyMojo extends AbstractHpiMojo {
             throw new MojoExecutionException("Cannot override self");
         }
 
-        Map<String, String> bundledPlugins = overrideWar != null ? scanWar(overrideWar, session, project) : Map.of();
+        Map<String, String> bundledPlugins = overrideWar != null ? scanWar(overrideWar) : Map.of();
         if (!bundledPlugins.isEmpty()) {
             getLog().info(String.format(
                     "Scanned contents of %s with %d bundled plugins",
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
```diff
@@ -517,8 +517,7 @@ private static void appendEntries(String property, Collection<String> additions,
      * @param war The WAR to scan.
      * @return The bundled plugins in the WAR.
      */
-    private static Map<String, String> scanWar(File war, MavenSession session, MavenProject project)
-            throws MojoExecutionException {
+    private Map<String, String> scanWar(File war) throws MojoExecutionException {
```
jtnord, line 520: `getLog()` is set for the mojo object, so this needs to be non-static.  the `MavenSession` and `MavenProject` are already set in the Mojo - so these are used directly.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
```diff
@@ -550,6 +549,12 @@ private static Map<String, String> scanWar(File war, MavenSession session, Maven
                         if (version == null) {
                             throw new IllegalArgumentException("Failed to determine version for " + name);
                         }
+                        // handle any extra info in snapshots e.g. " (private-abcd1234-username)"
```
jtnord, line 552: Whilst we could use `Implementation-Version` and then fall back - some plugins (`gradle`) do not provide this due to tooling or being too old (this was added approx. 2 years ago).  Given the need to fallback, it is easier just to handle the legacy case here.

src/main/java/org/jenkinsci/maven/plugins/hpi/TestDependencyMojo.java
```diff
@@ -550,6 +549,12 @@ private static Map<String, String> scanWar(File war, MavenSession session, Maven
...
+                        version = version.replaceFirst(" [(].+[)]$", "");
+                        if (version.endsWith("-SNAPSHOT")) {
+                            getLog().warn("War contains a SNAPSHOT of " + groupId + ":" + artifactId
+                                    + " build will not be fully repeatable");
```
basil, line 556: WAR is capitalized incorrectly, and there is a run-on sentence without punctuation:

```suggestion
                            getLog().warn("WAR contains a SNAPSHOT of " + groupId + ":" + artifactId
                                    + "; build will not be fully repeatable");
```

basil (approved):

PR #499, opened 2023-06-07
Remove use of Maven Wagon

basil:
Per [this page](https://cwiki.apache.org/confluence/display/MAVEN/Maven+Ecosystem+Cleanup) Maven Wagon is deprecated. This PR eliminates use of Maven Wagon at release time, replacing it with a GitHub Action that uploads the generated site to GitHub Pages on every clean build of the main branch.

### Testing done

I am using this workflow successfully in Job DSL plugin. If there are any issues with this PR I will fix them after merging this PR.

Fixes #492

PR #500, opened 2023-06-07
Restore mojo docs

basil:
Fixup to #499. Restores the mojo docs which inadvertently disappeared in that PR.

### Testing done

Ran `mvn -B -V -e -ntp -Pquick-build clean verify site` and verified the mojo docs were restored in the generated site.


basil:
http://jenkinsci.github.io/maven-hpi-plugin/hpi-mojo.html looks correct now

PR #514, opened 2023-07-04
Fix GH ribbon on site by updating maven-fluido-skin and keeping track of it

NotMyFault:
The current version of maven-fluido-skin pulls in a resource for AWS, which is no longer accessible, causing:

Screenshot 2023-07-04 at 16 37 14

The change proposed updates maven-fluido-skin to restore the ribbon, and keeps track of further updates:
Screenshot 2023-07-04 at 16 38 15


src/site/site.xml
```diff
@@ -10,14 +10,10 @@
   <publishDate format="yyyy-MM-dd" position="right" />
   <version position="right" />
 
-  <poweredBy>
-    <logo name="Built by Maven" href="http://maven.apache.org/" img="images/logos/maven-feather.png" />
-  </poweredBy>
-
```
NotMyFault, line 16: That's the default footer, which doesn't need to be specified separately.

timja (approved):

Issue #534, opened 2023-09-06
Integration Test failures related to the `pluginRepositories` order


dduportal:
Since https://github.com/jenkins-infra/jenkins-infra/pull/3041 was deployed to ci.jenkins.io, the Integration Tests are failing with this kind of errors:

```text
[INFO] Building: check-core-version-failure\pom.xml
[INFO] run post-build script verify.groovy
[INFO]   Assertion failed: 

assert new File(basedir, 'build.log').getText('UTF-8').contains("Dependency org.jenkins-ci.plugins:credentials:jar:1214.v1de940103927 requires Jenkins 2.375 or higher.")
       |        |                     |                |
       |        |                     |                false
[...]

PR #537, opened 2023-09-08
Restore build stability

dduportal:
This change fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/534 by allowing Integration Tests to execute in the context of ci.jenkins.io with its `settings.xml` mirroring setup.

This change is tested with and without a custom `settings.xml` to ensure both contributors and ci.jenkins.io environments are considered.

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,7 +24,7 @@ def runTests(Map params = [:]) {
             }
             // Needed for correct computation of JenkinsHome in RunMojo#execute.
             withEnv(['JENKINS_HOME=', 'HUDSON_HOME=']) {
-              infra.runMaven(args, params['jdk'])
+              infra.runMaven(args, params['jdk'], null, null, false)
             }
           }
         }
--- a/pom.xml
+++ b/pom.xml
@@ -246,6 +246,7 @@
           <cloneProjectsTo>${project.build.directory}/its</cloneProjectsTo>
           <localRepositoryPath>${basedir}/target/local-repo</localRepositoryPath>
           <settingsFile>src/it/settings.xml</settingsFile>
+          <mergeUserSettings>true</mergeUserSettings>
           <pomIncludes>
             <pomInclude>*</pomInclude>
           </pomIncludes>


dduportal:
I don't understand how and why this is failing on ci.jenkins.io. Even if I use the `settings.xml` from a ci.jenkins.io's AWS container agent, I can't reproduce the error (same temuring version but darwin/arm64 instead of linux/amd64, same Maven version). 

Gotta need more time to fully understand it (unless someone see the obvious I could have missed)

pom.xml
```diff
@@ -283,17 +283,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>mrm-maven-plugin</artifactId>
-        <version>1.5.0</version>
-        <configuration>
-          <propertyName>repository.proxy.url</propertyName>
```
basil, line 291: The `repository.proxy.url` a few lines above near line 250 can also be deleted now.

basil:
I believe that `pipeline-library` is setting `MAVEN_ARGS=-s <path-to-artifact-caching-proxy-settings>`. I believe this applies to the outermost Maven invocation (i.e., the one that is executed by the `Jenkinsfile` and runs `mvn clean install site […]`), but not the inner Maven invocations executed by `maven-invoker-plugin`, which define `settingsFile` to `src/it/settings.xml` which does not have the artifact caching proxy settings. Perhaps https://maven.apache.org/plugins/maven-invoker-plugin/integration-test-mojo.html#mergeUserSettings might be relevant here.

basil:
See also the documentation at https://maven.apache.org/plugins/maven-invoker-plugin/examples/fast-use.html#fast-build-configuration-mergeusersettings-and-mirrors

basil:
Even without the artifact caching proxying, trying to remove MRM fails as seen in #538, so I think we should keep MRM and focus this PR on simply restoring stability to the default branch as described in https://maven.apache.org/plugins/maven-invoker-plugin/examples/fast-use.html#fast-build-configuration-mergeusersettings-and-mirrors

dduportal:
> Even without the artifact caching proxying, trying to remove MRM fails as seen in #538, so I think we should keep MRM and focus this PR on simply restoring stability to the default branch as described in https://maven.apache.org/plugins/maven-invoker-plugin/examples/fast-use.html#fast-build-configuration-mergeusersettings-and-mirrors

Thanks for the help and the explanations. Your proposal make sense and it looks a good idea to focus on restoring build stability. 
I'm focusing on the settings.xml behavior you pointed, thanks for the pointers

dduportal:
Based on @basil's indications earlier in this thread, it looks like we (jenkins infra team) did not used the advanced Maven mirroring configuration properly.

Updating the default settings.xml of ci.jenkins.io in https://github.com/jenkins-infra/jenkins-infra/pull/3070

dduportal:
Test in progress in https://ci.jenkins.io/job/Plugins/job/maven-hpi-plugin/job/master/468/pipeline-graph/ for info, I'll check the result later

dduportal:
Update:
- Deployed to ci.jenkins.io the new `settings.xml` with the `external:` prefix in the `mirrorOf` property in https://github.com/jenkins-infra/jenkins-infra/commit/0d4b8e46c8aa6211c051bb1b6b9b873092d00501
- Build https://ci.jenkins.io/job/Plugins/job/maven-hpi-plugin/job/master/468/pipeline-graph/ (`master` branch of the `maven-hpi-plugin` failed with the same errors
- Added the commit https://github.com/jenkinsci/maven-hpi-plugin/pull/537/commits/0e10c8ff1e367755de9dad154bb3a37be77335a0 to merge the Maven settings: still failing (same error)
- I can't reproduce the problem on my machine which drives me crazy as I have the same `settings.xml` file as ci.jenkins.io. Trying to reproduce with the `-s` flag (instead of relying on the default `settings.xml`) to approach the merging method as close as ci.jenkins.io as possible
- In parallel, I've added https://github.com/jenkinsci/maven-hpi-plugin/pull/537/commits/3784afb37586287c4f5fd944dacc6cf1578e0d0a to remove the `mirrors` block from the IT's settings.xml , following the content of the page https://maven.apache.org/plugins/maven-invoker-plugin/examples/fast-use.html#fast-build-configuration. I'm under the impression that the `*` pattern of this section might take precedence on the ci.jenkins.io's directives

dduportal:
It seems I can reproduce the error locally by specifying the `-s <...>/settings.xml` option (instead of putting it in the default `$HOME/.m2/settings.xml`.

Most of the errors I have are unresolvable artefacts such as the plugin pom parent or the bom pom parent. I would have expected each IT run to download these artefacts when invoked by the inner Maven process but it seems to fail resolving these artifacts. I tried to remove the `-ntp` flag from the invoker properties for a given IT and checked the resulting log: the ACP I'm using is used for downloading a lot of files, but the failing artifacts are always using the `local.central` defined by the mrm. I might be missing something here.

I also wonder if another path would be to opt-out from the ACP for the `verify`phase, e.g. invoking the Maven `dependency:copy-dependencies` target first, in ci.jenkins.io, with the ACP enabled, and then running the `verify` phase without it 🤔 

Continuing diving this topic.

dduportal:
While I continue analyzing this topic (it drives me crazy, I have to understand the bottom of this!), I've pushed https://github.com/jenkinsci/maven-hpi-plugin/pull/537/commits/de66d6ee3a89715ec26c5fa75a9e2f38991eacb8 to test the behavior with a full opt-out from ACP.

If it works, I want to deliver it to make the Integration Tests running as expected on the principal branch to unblock contributors.

src/it/settings.xml
```diff
@@ -22,14 +22,6 @@ under the License.
 <settings xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.1.0 http://maven.apache.org/xsd/settings-1.1.0.xsd"
   xmlns="http://maven.apache.org/SETTINGS/1.1.0">
-  <mirrors>
-    <mirror>
-      <id>mrm-maven-plugin</id>
-      <name>Mock Repository Manager</name>
-      <url>@repository.proxy.url@</url>
-      <mirrorOf>*</mirrorOf>
-    </mirror>
```
basil, line 31: As mentioned previously, all of my efforts to remove `mrm-maven-plugin` here and in `plugin-pom` have failed, so I think `mrm-maven-plugin` should be retained.

dduportal: I wasn't sure if this block was mandatory: the documentation at https://maven.apache.org/plugins/maven-invoker-plugin/examples/fast-use.html#fast-build-configuration does not mention it.

I've removed this change from the PR, let's see

dduportal:
https://ci.jenkins.io/job/Plugins/job/maven-hpi-plugin/job/PR-537/12/ was successfull as far as I can tell (it is a replay of https://github.com/jenkinsci/maven-hpi-plugin/pull/537/commits/94f070db300527c98555616d3787bb12636f98c3 as I'm not a maintainer).

@basil what is your point of view about opting out of ACP?

Jenkinsfile
```diff
@@ -24,7 +24,7 @@ def runTests(Map params = [:]) {
             }
             // Needed for correct computation of JenkinsHome in RunMojo#execute.
             withEnv(['JENKINS_HOME=', 'HUDSON_HOME=']) {
-              infra.runMaven(args, params['jdk'])
+              infra.runMaven(args, params['jdk'], null, null, false)
```
basil, line 27: Does the build pass or fail with this hunk reverted?

dduportal: The build with only ACP disabled + the settings.xml merged passes (when replayed otherwise my Jenkinsfile change is not taken in account).

basil: Right. What about the build with this hunk reverted?

dduportal: I thought it failed but as I'm not so sure, let's retry. I've removed this hunk let's see

dduportal: I confirm that https://github.com/jenkinsci/maven-hpi-plugin/pull/537/commits/678094ac1aa1d40e85826bd0192db3c1ec25b226  fails (e.g. only "merging settings.xml" but no ACP).

dduportal: > I confirm that [678094a](https://github.com/jenkinsci/maven-hpi-plugin/commit/678094ac1aa1d40e85826bd0192db3c1ec25b226) fails (e.g. only "merging settings.xml" but no ACP) - build https://ci.jenkins.io/job/Plugins/job/maven-hpi-plugin/job/PR-537/14/

Hence my proposal: Add back the hunk which disable ACP for the build. It consumes a bit of bandwidth but ensures that Integration Tests are run without any problem.

basil (approved):
Thanks for the PR!

PR #538, opened 2023-09-14
Remove Mock Repository Manager Maven plugin

basil:
I cannot see a reason to continue using `mrm-maven-plugin`. Based on https://www.mojohaus.org/mrm/mrm-maven-plugin/examples/invoker-tests.html it seems the reason for using it was originally

> to allow the integration tests of a Maven Plugin to be run against a test local repository thereby ensuring that the plugin does not get installed into the local repository until after the integration tests have been confirmed as passing

but I just ran `mvn clean verify` and my local `~/.m2` repository did not seem to get polluted, so I don't think this is a concern anymore now that we are configuring `maven-invoker-plugin` with `localRepositoryPath`.

For what it's worth, I asked Maven developer Tamás Cservenák to audit `maven-hpi-plugin` a few months back, and here is what he had to say:

> I'd avoid the use of mrm plugin. You already pull Jetty, simplest (and most straightforward and least error prone) is to just fire up Jetty to serve files from somewhere as "remote repository" as in that case you are 100% in control of what is being served, while with MRM you are not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ mvn -f ../some-plugin -Dhpi-plugin.version=2.XXX-SNAPSHOT -DskipTests -DjenkinsH
 You can also rerun one test:
 
 ```bash
-mvn -Prun-its mrm:start invoker:run mrm:stop -Dinvoker.test=parent-4x
+mvn -Prun-its invoker:run -Dinvoker.test=parent-4x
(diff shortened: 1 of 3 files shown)


basil:
Removing MRM seems to have been a bad idea

PR #562, opened 2023-12-12
Push down parameters & methods

jglick:
Looking at sources I noticed that a lot of mojo parameters and helper methods were defined in an unnecessarily high superclass (used only in a specific subclass). Trying to push down things into more appropriate places.


jglick:
Abandoned; filed as https://github.com/jenkinsci/maven-hpi-plugin/issues/866 for posterity.

PR #578, opened 2024-02-07
Generate automatically a marker file for a maven profile depending on java level to use

Vlatombe:
(This idea came up as a discussion during the Jenkins contributor summit, based on the IDE problems experienced during the transition between Java 8 and 11 support)

This effectively reverts #403, reinstating #323, as well as adding a new mechanism for better IDE support.

The logic of setting `maven.compiler.release` dynamically works fine for the Maven CLI. However IDE support (at least IntelliJ) doesn't work well with this and happily ignores what the plugin set dynamically. Indeed, it has its own logic to parse the pom and apply profiles, but doesn't call any maven lifecycle doing so.

Instead, it picks up the default value set through the plugin pom that defaults to the minimum java version supported (currently [11](https://github.com/jenkinsci/plugin-pom/blob/6949f8af159d30e2722fcf773b397bc870e47359/pom.xml#L61)), instead of the version inferred from Java bytecode version obtained from Jenkins core.

Generating markerfiles allows to enable maven profiles that will set the java release to the correct version when loaded in IDE, as long as they are able to read activated profiles.

### Testing done
* Applied this patch
* Patched [plugin-pom](https://github.com/jenkinsci/plugin-pom) (see https://github.com/jenkinsci/plugin-pom/pull/893)
```
Subject: [PATCH] Add java level profiles
---
Index: pom.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/pom.xml b/pom.xml
--- a/pom.xml	(revision 6949f8af159d30e2722fcf773b397bc870e47359)
+++ b/pom.xml	(revision 28d462c55325e673d6f0c1710ac703300abae7b1)
@@ -1453,5 +1453,27 @@
         <yarn.lint.skip>true</yarn.lint.skip>
       </properties>
     </profile>
+    <profile>
+      <id>java-level-17</id>
+      <activation>
+        <file>
+          <exists>target/java-level/17</exists>
+        </file>
+      </activation>
+      <properties>
+        <maven.compiler.release>17</maven.compiler.release>
+      </properties>
+    </profile>
+    <profile>
+      <id>java-level-21</id>
+      <activation>
+        <file>
+          <exists>target/java-level/21</exists>
+        </file>
+      </activation>
+      <properties>
+        <maven.compiler.release>21</maven.compiler.release>
+      </properties>
+    </profile>
   </profiles>
 </project>
```
* built a jenkins snapshot with `-Dmaven.compiler.release=17`

Then in a plugin (I used `parallel-test-executor` as an example)
* Reference `maven-hpi-plugin` snapshot
* Reference `plugin-pom` snapshot
* Imported the project in IntelliJ (2024.1 EAP, they seem to have reworked the maven projects import, so it could matter)
* Check module settings -> Language level is set to 11 (expected)
* Update `jenkins.version` property to the snapshot built above
* Call `mvn validate`
* In IntelliJ, use **Maven > Reload project**
* Check module settings -> See the language level changing to 17 

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
@@ -1,0 +1,73 @@
+package org.jenkinsci.maven.plugins.hpi;
+
+import hudson.util.VersionNumber;
+import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+
+/**
+ * Configure Maven for the desired version of Java.
+ *
+ * @author Basil Crow
+ */
+@Mojo(name = "initialize", defaultPhase = LifecyclePhase.INITIALIZE)
+public class InitializeMojo extends AbstractJenkinsMojo {
+
+    @Override
+    public void execute() throws MojoExecutionException {
+        setCompilerProperties();
+    }
+
+    private void setCompilerProperties() throws MojoExecutionException {
+        if (!project.getProperties().containsKey("maven.compiler.source")
+                && !project.getProperties().containsKey("maven.compiler.release")) {
+            // On an older plugin parent POM that predates the setting of these values as Maven properties.
+            return;
+        }
+
+        JavaSpecificationVersion javaVersion = getMinimumJavaVersion();
+        if (JavaSpecificationVersion.forCurrentJVM().isOlderThan(new VersionNumber("9"))) {
+            // Should always be set already, but just in case...
+            setProperty("maven.compiler.source", javaVersion.toString());
+            setProperty("maven.compiler.target", javaVersion.toString());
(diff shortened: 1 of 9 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/InitializeMojo.java
```diff
@@ -0,0 +1,73 @@
+package org.jenkinsci.maven.plugins.hpi;
+
+import hudson.util.VersionNumber;
+import io.jenkins.lib.versionnumber.JavaSpecificationVersion;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugins.annotations.LifecyclePhase;
+import org.apache.maven.plugins.annotations.Mojo;
+
+/**
+ * Configure Maven for the desired version of Java.
```
jglick, line 10: Better to update to reflect its new more limited purpose.

jglick (approved):
I am not an IDEA user so I have no strong opinion. Has someone at least checked whether the IDEA bug is fixable in some reasonable way?

MarkEWaite (approved):

PR #582, opened 2024-02-21
`EnforceBytecodeVersion` fails with "Restricted to JDK 11 yet […] contains […] targeted to JDK 17"

basil:
End-to-end testing of Text Finder with a Jenkins core compiled with `-Dmaven.compiler.release=17` showed that #578 was incomplete: since the `initialize` logic ran after the Enforcer logic, the Enforcer logic still used the lowest supported Java version (not the automatically detected one) and failed. Fixed by moving the `initialize` logic to the `validate` file. This has the added benefit of making the code more readable by putting the (related) logic that writes the marker file and the logic that updates the value in-memory to an adjacent location in the same file.

### Testing done

Built a Jenkins core with `-Dmaven.compiler.release=17`, then ran `mvn clean verify` in Text Finder with that core and tip-of-trunk `maven-hpi-plugin` vs this PR. At tip-of-trunk, it failed with the Enforcer error shown above; with this PR it passes.


MarkEWaite (approved):
Thanks!

basil:
After testing this further, I see that this PR didn't actually fix the problem: I had forgotten to run `git clean -fxd` before testing this PR. The problem only occurs on a clean checkout: after the marker file is present, interpolation works as expected. So after writing the marker file, we not only need to update the in-memory settings for the Maven Compiler Plugin, but also we need to update them for Enforcer. I will look into this now.

PR #587, opened 2024-03-12
Generates `InjectedTest` in a named package.

Vlatombe:
This makes the test friendlier to JUnit results collection when working with PCT or in multi-module projects in general.

The package name is derived from groupId and artifactId, but can also be provided through configuration if needed.

### Testing done

Existing IT `parent-4x` exercises this, though doesn't make any assertion about the actual package name being used. Could add a new one if needed.

```[tasklist]
### Submitter checklist
```


basil (approved):
+1, seems reasonable to use a unique name for each plugin to facilitate test result visualizations.

jglick (approved):
Looks helpful. Should not break any tools which run `-Dtest=InjectedTest` since that will match a test inside a package.

src/test/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojoTest.java
```diff
@@ -0,0 +1,18 @@
...
+public class TestInsertionMojoTest {
+
+    @Test
+    public void legalizePackageName() {
+        assertEquals(
+                "org.jenkinsci.maven.plugins.hpi",
+                TestInsertionMojo.legalizePackageName("org.jenkinsci.maven.plugins.hpi"));
+        assertEquals(
+                "_123org.jenkinsci.maven.plugins.hpi",
+                TestInsertionMojo.legalizePackageName("123org.jenkinsci.maven.plugins.hpi"));
```
jglick, line 16: I suppose you might as well test `-` → `_` here.

jglick (commented):
I would suggest checking `javax.lang.model.SourceVersion.isName(CharSequence)` on the candidate package components and if not, either
* fixing it, like prepending `_`
* reporting a clear error and asking you to override the property
* falling back to putting the test in the default package

jglick (approved):

MarkEWaite (approved):

PR #699, opened 2024-12-13
Expand robustness of `isPlugin` against unbuildable models

jglick:
`isPlugin()` sometimes fails because the model for a dependency cannot currently be built. #104 had handled this in certain cases, but I just came across another case where https://github.com/jenkinsci/jenkins/pull/9958 meant that a dependency model could not be built because of

```
Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.apache.commons:commons-compress:jar is missing. @ line …, column …
```

when passing a newer `-Djenkins.version` from PCT: while `hpi:resolve-test-dependencies` worked as expected, `hpi:test-hpl` then failed because an old plugin dependency was being built against an older Jenkins core version and was picking up an implicit version of `commons-compress`.

### Testing done

For now, done manually using proprietary plugins; for reference, the problematic dep I tested `trigger-restrictions` → `operations-center-context`. Without the fix, the PCT run fails before even getting to `surefire:test`. With the fix (`-Dhpi-plugin.version=3.60-SNAPSHOT`) the PCT run passes.

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
@@ -1,7 +1,6 @@
 package org.jenkinsci.maven.plugins.hpi;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.LinkedHashSet;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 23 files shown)


raul-arabaolaza (approved):
LGTM, IIUC this only expands the existing logic for cases not covered before hence I see no risks here

jglick:
> this only expands the existing logic for cases not covered before hence I see no risks

I do not think it could break a previously working build. It _could_ make a build pass that (according to human judgment) should not have. The risk of even this seems low since it would have to involve a dependency on some artifact with a POM that could not be resolved (a relatively unusual event) but which has metadata similar to that of a Jenkins plugin yet which is not in fact a Jenkins plugin.

Vlatombe (approved):
Maybe worth an IT?

jglick:
Yeah, I just need to think up how to create one. Do not recall if there are already ITs that involve multiple plugins built from sources.

jglick:
I tried

<details><summary>this patch</summary>

```diff
diff --git src/it/override-test-dependencies-split-plugin/invoker.properties src/it/override-test-dependencies-split-plugin/invoker.properties
new file mode 100644
index 0000000..f470d07
--- /dev/null
+++ src/it/override-test-dependencies-split-plugin/invoker.properties
@@ -0,0 +1 @@
+invoker.goals=-ntp -DoverrideVersions=org.jenkins-ci.plugins:jobcacher:573.v33fa_12644a_91 -Djenkins.version=2.489 test
diff --git src/it/override-test-dependencies-split-plugin/pom.xml src/it/override-test-dependencies-split-plugin/pom.xml
new file mode 100644
index 0000000..b7f4baa
--- /dev/null
+++ src/it/override-test-dependencies-split-plugin/pom.xml
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.jenkins-ci.plugins</groupId>
+    <artifactId>plugin</artifactId>
+    <version>4.86</version>
+    <relativePath/>
+  </parent>
+  <groupId>org.jenkins-ci.tools.hpi.its</groupId>
+  <artifactId>override-test-dependencies-smokes</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <packaging>hpi</packaging>
+  <properties>
+    <jenkins.version>2.452.4</jenkins.version>
+    <hpi-plugin.version>@project.version@</hpi-plugin.version>
+    <test>SampleTest</test>
+    <maven.test.redirectTestOutputToFile>false</maven.test.redirectTestOutputToFile>
+    <spotless.check.skip>false</spotless.check.skip>
+  </properties>
+  <dependencies>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>jobcacher</artifactId>
+      <version>551.ve0b_00cb_1b_85c</version>
+    </dependency>
+  </dependencies>
+  <repositories>
+    <repository>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
+    </repository>
+  </repositories>
+  <pluginRepositories>
+    <pluginRepository>
+      <id>repo.jenkins-ci.org</id>
+      <url>https://repo.jenkins-ci.org/public/</url>
+    </pluginRepository>
+  </pluginRepositories>
+</project>
diff --git src/it/override-test-dependencies-split-plugin/src/main/resources/index.jelly src/it/override-test-dependencies-split-plugin/src/main/resources/index.jelly
new file mode 100644
index 0000000..2f655e5
--- /dev/null
+++ src/it/override-test-dependencies-split-plugin/src/main/resources/index.jelly
@@ -0,0 +1,2 @@
+<?jelly escape-by-default='true'?>
+<div/>
diff --git src/it/override-test-dependencies-split-plugin/src/test/java/sample/SampleTest.java src/it/override-test-dependencies-split-plugin/src/test/java/sample/SampleTest.java
new file mode 100644
index 0000000..7cdfae9
--- /dev/null
+++ src/it/override-test-dependencies-split-plugin/src/test/java/sample/SampleTest.java
@@ -0,0 +1,13 @@
+package sample;
+
+import jenkins.plugins.jobcacher.ArbitraryFileCache;
+import org.junit.Test;
+
+public class SampleTest {
+
+    @Test
+    public void smokes() throws Exception {
+        // just load some org.apache.commons.compress types:
+        ArbitraryFileCache.CompressionMethod.TARGZ.getCacheStrategy();
+    }
+}
```

</details> 

but it was no good (passes even in trunk). The problem is that the change made to a dependency plugin which required this was in a proprietary plugin, replacing

```xml
<dependency>
  <groupId>org.apache.commons</groupId>
  <artifactId>commons-compress</artifactId>
  <scope>provided</scope>
</dependency>
```

with

```xml
<dependency>
  <groupId>io.jenkins.plugins</groupId>
  <artifactId>commons-compress-api</artifactId>
</dependency>
```

But the only OSS plugin PR I could find linked to the core PR that had a similar pattern, https://github.com/jenkinsci/jobcacher-plugin/pull/357, was also using Incrementals, and so the flattened POM already had

```xml
<dependency>
  <groupId>org.apache.commons</groupId>
  <artifactId>commons-compress</artifactId>
  <version>1.26.0</version>
  <scope>compile</scope>
</dependency>
```

which does not need to refer to Jenkins core to resolve.

Nor could I see how to create such a fake dependency plugin, `install` two versions of it, then run a separate `mvn test` command against the IT plugin referring to those two versions; there are a few tests that use multimodule, but in those cases the `<modules>` are in the same reactor, which would go through a totally different model resolution process.

jglick:
https://maven.apache.org/plugins/maven-invoker-plugin/examples/prepare-build-env.html :eyes:

jglick:
There we go: a [failure](https://github.com/jenkinsci/maven-hpi-plugin/pull/701/checks?check_run_id=34614835800)

```
Failed to execute goal org.jenkins-ci.tools:maven-hpi-plugin:3.60-rc1663.f9dcd86e2750:test-hpl (default-test-hpl) on project override-test-dependencies-smokes: Error preparing the manifest: Failed to open artifact io.jenkins.plugins:somedep:jar:1:compile at /home/jenkins/agent/workspace/Plugins_maven-hpi-plugin_PR-701/target/local-repo/io/jenkins/plugins/somedep/1/somedep-1.jar: org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for org.apache.commons:commons-compress:jar is missing. @ line 18, column 17: 1 problem was encountered while building the effective model for io.jenkins.plugins:somedep:1
[ERROR] 'dependencies.dependency.version' for org.apache.commons:commons-compress:jar is missing. @ line 18, column 17
```

src/main/java/org/jenkinsci/maven/plugins/hpi/WarMojo.java
```diff
@@ -54,58 +53,54 @@ public class WarMojo extends RunMojo {
      */
     @Override
     public void execute() throws MojoExecutionException {
-        try {
-            if (outputFile == null) {
-                outputFile = new File(
-                        getProject().getBasedir(), "target/" + getProject().getArtifactId() + ".war");
-            }
+        if (outputFile == null) {
```
jglick, line 56: (ignore WS)

Vlatombe (approved):

PR #702, opened 2024-12-19
[Jenkins-65548] Fix - Random ordering of generated sources

code-arnab:
### Testing done
Closes this - [JIRA Ticket](https://issues.jenkins.io/browse/JENKINS-65548?jql=labels%20%3D%20newbie-friendly%20AND%20status%20not%20in%20(Closed%2C%20Done%2C%20Resolved%2C%20%22Fixed%20but%20Unreleased%22%2C%20%22In%20Review%22%2C%20%22In%20Progress%22)%20AND%20updatedDate%20%3E%202020-01-01%20AND%20component%20%3D%20maven-hpi-plugin%20AND%20project%20%3D%20JENKINS%20AND%20priority%20%3D%20Major%20ORDER%20BY%20updated%20DESC)

To fix this issue, I've utilized Arrays.sort() (Alphabetically sorted by ignoring cases). This aims to maintain a consistent order which is necessary for reproducible builds. 

### Submitter checklist
- [x] Link to relevant pull requests, esp. upstream and downstream changes


src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
```diff
@@ -118,6 +119,10 @@ private void report(JPackage pkg, String fileName) {
     private void walk(File dir, JPackage pkg, String dirName) throws JClassAlreadyExistsException, IOException {
         File[] children = dir.listFiles(File::isDirectory);
         if (children != null) {
+            // Sorting the children to maintain a consistent order
+            // This is important for reproducible builds
+            Arrays.sort(children, (a, b) -> a.getName().compareToIgnoreCase(b.getName()));
```
jglick, line 124: Generally prefer simple case-sensitive compare unless there is a specific reason to make it case-insensitive. I guess it should not matter here because the only way the current code could result in an unreproducible order is if the directory contained two files differing only in case, like `something.jelly` + `someThing.jelly`, in which case the random order from `listFiles` would be preserved by this `sort`. But in that situation the repository could not even be checked out on Windows so someone would probably have noticed the mistake long ago anyway.

jglick (approved):

src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
code-arnab: Thank You for the tip. Will keep in mind.

PR #803, opened 2025-10-24
Support dots in Jelly file names

janfaracik:
Experimenting with a component at the moment, titled `defer.placeholder`. Having a `.` is invalid at the moment when converted to Java - so I've updated this to rename the dot to an underscore (same way it handled hyphens).

### Testing done

* Jenkins builds successfully with this change (previously it didn't) for https://github.com/jenkinsci/jenkins/pull/11222

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/TagLibInterfaceGeneratorMojo.java
@@ -158,7 +158,7 @@ public class TagLibInterfaceGeneratorMojo extends AbstractMojo {
                         String baseName = FilenameUtils.getBaseName(tag.getName());
                         String methodName;
                         if (!JJavaName.isJavaIdentifier(tag.getName())) {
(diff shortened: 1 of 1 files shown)


daniel-beck:
What is the benefit of `foo.bar` over `foo-bar`? Just that it's an arbitrary naming scheme used elsewhere?

timja (approved):
Seems fine to me, @daniel-beck any objection to it? Or just understanding justification.

I did some research on it in general, XAML uses a dot for it for it to semantically associate to the parent. Regular XML doesn't care about it at all.

https://stackoverflow.com/a/1550126/4951015

timja:
@daniel-beck are you ok with proceeding? (If we don't hear back from Tuesday I intend to merge this)

daniel-beck:
No specific objections at this point, but there seems to be no real motivation, not even a linked PR. Is there test coverage for this, beyond just a tag being used in Jelly? (E.g., does this work with Groovy views? How's the generated Jellydoc? I assume these work, but confirmation would be good.)

PR #814, opened 2025-11-16
Add property for detection of inline javascript

mawinter69:
Add property `jelly.prohibitInlineJS`. This is passed to the injected tests as parameters.
When https://github.com/jenkinsci/jenkins-test-harness/pull/868 gets merged this will enable the checks.
By default the checks are enabled here.
At a later point in time we might consider removing this property so that it is no longer possible to disable the check. Or we might ignore it when the Jenkins core version is higher than a certain version.

### Testing done
Tested that in combination with https://github.com/jenkinsci/jenkins-test-harness/pull/868 setting the property to false will not trigger the check.

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,20 +56,6 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>org.eclipse.jetty</groupId>
-        <artifactId>jetty-bom</artifactId>
-        <version>${jetty.version}</version>
-        <type>pom</type>
-        <scope>import</scope>
-      </dependency>
-      <dependency>
-        <groupId>org.eclipse.jetty.ee9</groupId>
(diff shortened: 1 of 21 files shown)


src/main/java/org/jenkinsci/maven/plugins/hpi/TestInsertionMojo.java
```diff
@@ -68,6 +69,12 @@ public class TestInsertionMojo extends AbstractJenkinsMojo {
     @Parameter(property = "jelly.requirePI", defaultValue = "true")
     private boolean requirePI;
 
+    /**
+     * If true, verify that all the jelly scripts have the Jelly XSS PI in them.
```
daniel-beck: Bad copy & paste

NotMyFault (approved):
Untested, but code looks reasonable.

PR #862, opened 2026-01-18
Fork jenkins with winstone instead of jetty plugin

timja:
### **Replace Jetty-based `hpi:run` with a Winstone-based launcher**

Production-like dev environment, run with same setup as users. No more missing configuration done in winstone like in https://github.com/jenkinsci/maven-hpi-plugin/issues/625
  - No more classpath conflicts between maven classes and Jenkins classes: https://github.com/jenkinsci/maven-hpi-plugin/issues/818
  - Pipelines work on Java 25 - mentioned in https://github.com/jenkinsci/maven-hpi-plugin/issues/840
  - Removed `sun.misc.Unsafe` usage which resolves warnings when run on Java 25 about a terminally deprecated field
  - No longer need to keep Jetty in-sync here as `jenkins.version` will manage that and ensure core compatibility with jetty versions
  - Adds support for `mvnd` with `hpi:run` - https://github.com/jenkinsci/maven-hpi-plugin/issues/839

### **Debugging changes**

Adds `-Dmaven.hpi.debug` option to control debugging of the forked JVM:
- `"true"` (or if a debugger is already attached to Maven) enables JDWP with `suspend=y` on a random port.
- Otherwise allows passing a custom debug string directly into the JVM args.

This _should_ be the only behaviour change with this pull request. As the Jenkins process is no longer the same as the maven process you can't just click debug in the IDE and have it work.

If you do click debug in the IDE it will detect that there is a debugger present and create a new debug session in the forked process, this will be on a random port (to avoid conflicts with multiple `hpi:run` instances).

IntelliJ has a link you can click to attach to the process:
image

Then debugging will work as normal.

Otherwise you can configure your own args in `-Dmaven.hpi.debug`.
There is also an IntelliJ plugin that might work (untested): https://plugins.jetbrains.com/plugin/13263-attachme 

Fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/840
Fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/625
Fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/818
Fixes https://github.com/jenkinsci/maven-hpi-plugin/issues/839

- https://github.com/jenkinsci/maven-hpi-plugin/issues/840
- https://github.com/jenkinsci/maven-hpi-plugin/issues/625
- https://github.com/jenkinsci/maven-hpi-plugin/issues/818
- https://github.com/jenkinsci/maven-hpi-plugin/issues/839

### Testing done

[...]

PR #863, opened 2026-01-24
Default debug port to 5005

timja:
see https://github.com/jenkinsci/maven-hpi-plugin/pull/862#discussion_r2724152881

Why 5005? Its the default port for IntelliJ debug run configurations, so no need to change it when you create a remote debug configuration.

### Testing done

Checked it starts on 5005 and that I can customise it

### Submitter checklist

### DIFF
--- a/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
+++ b/src/main/java/org/jenkinsci/maven/plugins/hpi/RunMojo.java
@@ -122,6 +122,14 @@ public class RunMojo extends AbstractHpiMojo {
     private String debugForkedProcess;
 
     /**
+     * Port number for the debugger to attach to.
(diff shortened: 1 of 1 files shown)


jglick:
https://github.com/jenkinsci/maven-hpi-plugin/pull/862 seems to have changed how debugging worked. NetBeans would run e.g.

```bash
MAVEN_OPTS=-agentlib:jdwp=transport=dt_socket,server=n,address=45297 mvn -Djpda.listen=maven -Djpda.address=45297 hpi:run
```

with some randomly allocated port. This no longer works; I have to attach to port 5005 after it starts. Also debugging does not work at all when using `mvnd` AFAICT. Seems like this is an area that could be refined though I am not exactly sure how yet.

